PR #3, opened 2022-05-17
Drop NVreg_EnablePCIeGen3=1

ventureoo:
Please accept it, for I do not see the point of this parameter at this time. @ptr1337


ptr1337 (approved):
Thanks for checking! approved

ptr1337:
Thanks for your PR @ventureoo 

It would be nice if you start signing also your commits :)

ventureoo:
Yeah, that would be nice.

PR #9, opened 2023-01-26
Update-network-config

SoongVilda:
Added values from blog post https://blog.cloudflare.com/optimizing-tcp-for-high-throughput-and-low-latency/

I tested these values, and I've got smoother experiences with NVIDIA GeForce Now, which also helped me decrease streaming latency on LAN.


ptr1337:
These settings need to be tested.
I would include the cloudflare patch with the next kernel version and then adding its options.

ptr1337:
Required patch is queued for 6.1.9. Will be tested as soon released.

PR #11, opened 2023-01-29
Add udev rules for proper working of dynamic power management on NVIDIA

ventureoo:
According to NVIDIA documentation: https://download.nvidia.com/XFree86/Linux-x86_64/525.85.05/README/dynamicpowermanagement.html

Extremely demanding to test.

### DIFF
--- a/etc/udev/rules.d/80-nvidia-pm.rules
+++ b/etc/udev/rules.d/80-nvidia-pm.rules
@@ -1,0 +1,7 @@
+# Enable runtime PM for NVIDIA VGA/3D controller devices on driver bind
+ACTION=="bind", SUBSYSTEM=="pci", ATTR{vendor}=="0x10de", ATTR{class}=="0x030000", TEST=="power/control", ATTR{power/control}="auto"
+ACTION=="bind", SUBSYSTEM=="pci", ATTR{vendor}=="0x10de", ATTR{class}=="0x030200", TEST=="power/control", ATTR{power/control}="auto"
+
+# Disable runtime PM for NVIDIA VGA/3D controller devices on driver unbind
+ACTION=="unbind", SUBSYSTEM=="pci", ATTR{vendor}=="0x10de", ATTR{class}=="0x030000", TEST=="power/control", ATTR{power/control}="on"
+ACTION=="unbind", SUBSYSTEM=="pci", ATTR{vendor}=="0x10de", ATTR{class}=="0x030200", TEST=="power/control", ATTR{power/control}="on"
--- a/etc/udev/rules.d/80-nvidia-pm.rules
+++ b/etc/udev/rules.d/80-nvidia-pm.rules
@@ -1,7 +1,5 @@
 # Enable runtime PM for NVIDIA VGA/3D controller devices on driver bind
-ACTION=="bind", SUBSYSTEM=="pci", DRIVERS=="nvidia", ATTR{vendor}=="0x10de", ATTR{class}=="0x030000", TEST=="power/control", ATTR{power/control}="auto"
-ACTION=="bind", SUBSYSTEM=="pci", DRIVERS=="nvidia", ATTR{vendor}=="0x10de", ATTR{class}=="0x030200", TEST=="power/control", ATTR{power/control}="auto"
+ACTION=="bind", SUBSYSTEM=="pci", DRIVERS=="nvidia", ATTR{vendor}=="0x10de", ATTR{class}=="0x03[0-9]*", TEST=="power/control", ATTR{power/control}="auto"
 
 # Disable runtime PM for NVIDIA VGA/3D controller devices on driver unbind
-ACTION=="unbind", SUBSYSTEM=="pci", DRIVERS=="nvidia", ATTR{vendor}=="0x10de", ATTR{class}=="0x030000", TEST=="power/control", ATTR{power/control}="on"
-ACTION=="unbind", SUBSYSTEM=="pci", DRIVERS=="nvidia", ATTR{vendor}=="0x10de", ATTR{class}=="0x030200", TEST=="power/control", ATTR{power/control}="on"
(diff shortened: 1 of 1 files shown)


ventureoo:
I was suddenly worried about how these rules would behave with the nouveau driver. We need to be sure of that.

SoongVilda:
@ventureoo I believe we don't need worry about nouveau driver, because you can select "latest nvidia card" before CachyOS boot up. Anyway, I think we should make that as part of mhwd.

If nvidia card, then apply these nvidia udev rules. Then we can also add HW acceleration video encoding with NVENC.
https://wiki.archlinux.org/title/NVIDIA#Hardware_accelerated_video_encoding_with_NVENC

ventureoo:
@SoongVilda I added an extra check so that the rule is applied only if the NVIDIA module is loaded. So I think everything is fine.

ventureoo:
> If nvidia card, then apply these nvidia udev rules. Then we can also add HW acceleration video encoding with NVENC.
https://wiki.archlinux.org/title/NVIDIA#Hardware_accelerated_video_encoding_with_NVENC

Do we really need these rules when nvidia-utils already includes them or is that not it?
https://github.com/archlinux/svntogit-packages/blob/packages/nvidia-utils/trunk/nvidia.rules

SoongVilda:
Seems to be good for me now, let's wait for Ptr to comment that. 

Anyway, what do you think about this?
https://github.com/Frogging-Family/nvidia-all/blob/master/60-nvidia.rules

ventureoo:
> Anyway, what do you think about this? https://github.com/Frogging-Family/nvidia-all/blob/master/60-nvidia.rules

I was looking at the rules from nvidia-all. The first part repeats what nvidia-utils already has. The second part is weirder, because I didn't see any mention in the driver documentation about enabling runtime PM also for USB/Audio devices.
The negativo17 driver does not have anything like that either.
https://github.com/negativo17/nvidia-kmod-common/blob/master/60-nvidia.rules

ptr1337:
> I was suddenly worried about how these rules would behave with the nouveau driver. We need to be sure of that.

We need to look into this.
We have currently a issue with the latest CachyOS ISO, that nouveau does not get proper loaded and people need to start it with  `modprobe nouveau`. I dont know why this is happening.
First I thought this would be due the CachyOS-Settings package, but I made a ISO without CachyOS-Settings and the User did still had this issue.
Since we package the nvidia driver with, there could be a issue in the recent changes from them which blocks to load the nouveau driver automatically, cause this issue was not present in the ISO 3 months ago.



> Seems to be good for me now, let's wait for Ptr to comment that.
> 
> Anyway, what do you think about this? https://github.com/Frogging-Family/nvidia-all/blob/master/60-nvidia.rules

nvidia-all does have a lot of "legacy" stuff in and it seems not to be proper maintained.

ventureoo:
>  We have currently a issue with the latest CachyOS ISO, that nouveau does not get proper loaded and people need to start it with modprobe nouveau. I dont know why this is happening.

I think this might be a mhwd issue?

Anyway, I think with https://github.com/CachyOS/CachyOS-Settings/pull/11/commits/a4eea870b7fbc1ee2300061079aeacf490dc37bb it shouldn't be such a big problem. However, we should also check how the old 390xx drivers react to these rules.

ptr1337:
> > We have currently a issue with the latest CachyOS ISO, that nouveau does not get proper loaded and people need to start it with modprobe nouveau. I dont know why this is happening.
> 
> I think this might be a mhwd issue?
> 
> Anyway, I think with [a4eea87](https://github.com/CachyOS/CachyOS-Settings/commit/a4eea870b7fbc1ee2300061079aeacf490dc37bb) it shouldn't be such a big problem. However, we should also check how the old 390xx drivers react to these rules.

No, MHWD is not active at the ISO itself. But we need to watch into.

From my side is fine, I have tested it on nvidia

ptr1337 (approved):

vnepogodin:
Mhwd is only used in the installer as one of last steps.
It does not get loaded or download anything during the boot

NextWork123 (changes requested):
Hello, you can modify the third line to: `ACTION=="bind", SUBSYSTEM=="pci", ATTR{class}=="0x030200", TEST=="power/control", ATTR{power/control}="auto"`

And modify the seventh line to: `ACTION=="unbind", SUBSYSTEM=="pci", ATTR{class}=="0x030200", TEST=="power/control", ATTR{power/control}="on"`

There's no need to specify DRIVERS or ATTR{vendor} again, as you have already done so in the first and sixth lines.

ventureoo:
> Hello, you can modify the third line to: `ACTION=="bind", SUBSYSTEM=="pci", ATTR{class}=="0x030200", TEST=="power/control", ATTR{power/control}="auto"`
> 
> And modify the seventh line to: `ACTION=="unbind", SUBSYSTEM=="pci", ATTR{class}=="0x030200", TEST=="power/control", ATTR{power/control}="on"`
> 
> There's no need to specify DRIVERS or ATTR{vendor} again, as you have already done on the first and sixth lines.

I think we can just replace the class by a pattern, as it is done in nvidia-all.

NextWork123:
> > Hello, you can modify the third line to: `ACTION=="bind", SUBSYSTEM=="pci", ATTR{class}=="0x030200", TEST=="power/control", ATTR{power/control}="auto"`
> > And modify the seventh line to: `ACTION=="unbind", SUBSYSTEM=="pci", ATTR{class}=="0x030200", TEST=="power/control", ATTR{power/control}="on"`
> > There's no need to specify DRIVERS or ATTR{vendor} again, as you have already done on the first and sixth lines.
> 
> I think we can just replace the class by a pattern, as it is done in nvidia-all.

Yes, you can do that too

NextWork123 (approved):
LGTM

SoongVilda:
@ptr1337 Can I merge that?

ptr1337:
Yes

PR #15, opened 2023-03-20
Add rules to prevent of lacks nvidia modules

ventureoo:
Basically just a backport of rules from Ubuntu, but this should fix us from being tied to mhwd-gpu. Should have no effect on non-NVIDIA configurations, including nouveau.

P. S. I'm not sure if I should merge them with the 80-nvidia-pm.rules, so I made a separate file.


SM9CC:
Looks good to me, my suggestion would be to merge all of the nvidia settings into one file so we just have 

> 71-nvidia.rules

with everything, other than that suggestion, I see no problems merging this.

ventureoo:
> Looks good to me, my suggestion would be to merge all of the nvidia settings into one file so we just have
> 
> > 71-nvidia.rules
> 
> with everything, other than that suggestion, I see no problems merging this.

Done.

PR #16, opened 2023-04-06
Moving wayland variables from shell configs to /etc/profile.d/cachyos-wayland.sh

norz3n:
Greetings!

I propose to move the variables for Wayland to profile.d so that they work regardless of the default command line shell. Also, profile.d itself will allow you to make interesting tweaks to the system's variables.

A little later I will make pull requests for the zsh and fish settings, mentioning this pull request


NextWork123:
Hello, After a review of your code i think isn't good to do because fish shell isn't POSIX compatible so it will break all. 

I will flag that as "bad idea" and closing that pull

Best Regards,
NextWorks

PR #17, opened 2023-04-19
pci-latency: Replace grep with the lspci device filter by class

ventureoo:
Just worried about possible errors because of grep


sm9cc:
Thanks!

PR #18, opened 2023-04-23
sysctl: Disable split_lock_mitigate by default

ventureoo:
Since 6.2 this behavior can be controlled via sysctl instead of a kernel parameter, so use that, although it will not work for LTS.

See also: https://www.phoronix.com/news/Linux-Splitlock-Hurts-Gaming

### DIFF
--- a/etc/sysctl.d/99-cachyos-settings.conf
+++ b/etc/sysctl.d/99-cachyos-settings.conf
@@ -134,3 +134,7 @@ kernel.sched_pelt_multiplier = 2
 # Increase the sched_rt_runtime_us to mitigate issues:
 # sched: RT throttling activated
 kernel.sched_rt_runtime_us=980000
+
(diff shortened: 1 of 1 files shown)


ptr1337:
> Since 6.2 this behavior can be controlled via sysctl instead of a kernel parameter, so use that, although it will not work for LTS.
> 
> See also: https://www.phoronix.com/news/Linux-Splitlock-Hurts-Gaming

Im against this. Actually this setting is really related to one game. Maybe more.
This is a security mechanism and we have a entry for this in the wiki, so if people face this issue they can change it.

ptr1337:
https://wiki.cachyos.org/en/home/General_System_Tweaks

Mentioned here

ventureoo:
> https://wiki.cachyos.org/en/home/General_System_Tweaks
> 
> Mentioned here

Sorry, I missed that.

PR #21, opened 2023-05-07
Clean up sysctl

ventureoo:
The settings for the network can be reviewed and returned. But for now, they will be removed.

Also note what the Arch Wiki says about these settings:
> This section seems motivated by the Cloudflare blog post, but Red Hat's tuned profile suggests even smaller values and claims "this seems only necessary at 40Gb speeds". Hence, these settings do not seem useful for commodity hardware.

https://wiki.archlinux.org/title/Sysctl#Increase_the_memory_dedicated_to_the_network_interfaces
https://github.com/redhat-performance/tuned/blob/master/profiles/network-throughput/tuned.conf#L10

I also have some doubts about the current values of dirty_ratio.

### DIFF
--- a/etc/sysctl.d/99-cachyos-settings.conf
+++ b/etc/sysctl.d/99-cachyos-settings.conf
@@ -63,22 +63,6 @@ kernel.kexec_load_disabled = 1
 # The upper limit on how many connections the kernel will accept (default 128): 
 net.core.somaxconn = 8192
 
-# Increase the memory dedicated to the network interfaces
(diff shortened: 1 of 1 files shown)


ptr1337 (approved):
Seems good to me.
Actually im not sure increasing the vm.dirty_ratio, since this generally could be set with a script depending on the Memory of the users system. But increasing it as default is a good point.

Thanks for this update.

Maybe @jomo can help something related to this script.

sirlucjan (approved):

sirlucjan:
LGTM

PR #25, opened 2023-05-21
modprobe.d: Force using of amdgpu driver for older graphics cards by default

ventureoo:
We have already received a lot of issues from users about Vulkan not working on old AMD cards due to lack of amdgpu driver. I think it makes sense to force its use by default, since users will want to do it anyway and our kernel supports this feature. 

Refs:
https://wiki.archlinux.org/title/AMDGPU#Enable_Southern_Islands_(SI)_and_Sea_Islands_(CIK)_support
https://github.com/ValveSoftware/Proton/wiki/For-AMD-users-having-issues-with-non-OpenGL-games
https://gitlab.freedesktop.org/drm/amd/-/issues/1776

### DIFF
--- a/etc/modprobe.d/amdgpu.conf
+++ b/etc/modprobe.d/amdgpu.conf
@@ -3,6 +3,5 @@
 options amdgpu si_support=1
 options amdgpu cik_support=1
 
-# If amdgpu fails to boot, will do fallback on radeon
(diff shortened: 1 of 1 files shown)


SoongVilda:
@ventureoo Is that everything for now? Can we merge?

PR #26, opened 2023-05-28
Update 99-cachyos-settings.conf

SoongVilda:
Possible Fix from Array


SoongVilda:
I badly understand to fix, I going to close that.

PR #31, opened 2023-08-31
Enable defer+madvise defragmention for THP by default

ventureoo:
This should improve the performance of applications that use the tcmalloc memory allocator, which is used by Chromium/Electron and Valve's native games, among others.
Ref: https://github.com/google/tcmalloc/blob/master/docs/tuning.md#system-level-optimizations

### DIFF
--- a/etc/tmpfiles.d/thp.conf
+++ b/etc/tmpfiles.d/thp.conf
@@ -1,4 +1,3 @@
 # Improve performance for applications that use tcmalloc
 # https://github.com/google/tcmalloc/blob/master/docs/tuning.md#system-level-optimizations
 w! /sys/kernel/mm/transparent_hugepage/defrag - - - - defer+madvise
-w! /sys/kernel/mm/transparent_hugepage/enabled - - - - madvise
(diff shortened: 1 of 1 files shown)


Boria138:
I guess the settings could be a little more customized

w- /sys/kernel/mm/transparent_hugepage/enabled - - - - - madvise
w- /sys/kernel/mm/transparent_hugepage/khugepaged/defrag - - - - - - 1
w- /sys/kernel/mm/transparent_hugepage/defrag - - - - - defer+madvise

Boria138:
@ventureoo 
```Failed to write file "/sys/kernel/mm/transparent_hugepage/defrag": Invalid argument```

Torston420:
[Kernel docs](https://www.kernel.org/doc/html/next/admin-guide/mm/transhuge.html#global-thp-controls) imply that 'madvise' would perform the stalling behavior seen in 'always', wouldn't 'defer' be more advantageous than 'defer+madvise' for interactivity?

ventureoo:
> Update: Team Fortress 2 does not seem to allocate any hugepages at all if /transparent_hugepage/enabled is set to 'madvise' instead of 'always', this appears to be a global issue with TCMalloc: https://github.com/google/tcmalloc/issues/85

Thanks for notice. I'll remove madvise for `transparent_hugepage/enabled` in that case.

ventureoo:
@ptr1337 I've signed the last commit. Are there any objections to this PR?

PR #36, opened 2023-09-22
Update zram-generator.conf to use LZ4

rialpa84:
LZ4 is what is configured to be used in /etc/udev/rules.d/30-zram.rules so this is just to sync up with that.

### DIFF
--- a/etc/systemd/zram-generator.conf
+++ b/etc/systemd/zram-generator.conf
@@ -1,3 +1,3 @@
 [zram0] 
-compression-algorithm = zstd
+compression-algorithm = lz4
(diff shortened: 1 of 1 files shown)


rialpa84:
May I ask why Zstd-3 is the default?

It's slower than LZ4 for write speeds and if a person has a lot of memory, 64GB in my case, then the default 16GB allotted towards Zram using Zstd-3 for a higher compression ratio is moot and unnecessary. If a person has 8GB of ram, sure, Zstd is great to assist a 2GB Zram, but when you have a lot of ram the higher write performance of LZ4 should take precedence.

PR #37, opened 2023-09-23
tcp patches

Rasenkai:


etc/sysctl.d/99-cachyos-settings.conf
```diff
@@ -74,6 +74,21 @@ net.ipv4.tcp_congestion_control = bbr
 # Helps protect against SYN flood attacks. Only kicks in when net.ipv4.tcp_max_syn_backlog is reached: 
 net.ipv4.tcp_syncookies = 1
 
+# TCP Enable ECN Negotiation by default
+net.ipv4.sysctl_tcp_ecn = 1
```
Torston420, line 78: Are you sure you didn't mean to write `net.ipv4.tcp_ecn`?

ventureoo:
We had previously experienced some regressions with sysctl parameters for the network, so we gave up on them. But this set doesn't seem to be that large, so we can test this and then decide which parameters are okay. I think @array is more compotent on the question, which is why I listed him as a reviewer.

aarrayy:
Should be fine but afaik TCP timestamping is needed for some of the BBRv3 functionality, if you take a look into their source code, they're some references about timestamps and RTT calculations and if i'm not wrong net.ipv4.tcp_timestamps = 0 could make it worse?, besides this. The other three are fine and could be added

aarrayy (commented):
This could affect BBRv3

aarrayy (approved):

ptr1337:
Please sign your commits @Rasenkai

PR #41, opened 2023-11-11
nvidia: Add fbdev=1 as default

ptr1337:
This adds the new nvidia option "fbdev=1" as default for the nvidia module.
This option is required for some nvidia cards to have a proper wayland experience. [1}

On top this provides a framebuffer, which is scaled to the default monitor resolution.

1. https://forums.developer.nvidia.com/t/kde-wayland-doesnt-work-on-the-545-drivers/269794/2


ventureoo (approved):
LGTM so far, but if there are any regressions, we could revert this.

Issue #34, opened 2023-09-17
New IO scheduler settings


John-Gee:
Hello,

I wondered if it would be good to investigate new IO settings based on: https://github.com/pop-os/default-settings/pull/149
While none is often recommended for nvme drives, is it really good on interactive devices or just on servers?

John-Gee:
Do you know of any interesting happening around K2? https://github.com/TUD-OS/k2-scheduler/

ptr1337:
Hi @John-Gee 

I have made a PR after 2 months of testing "kyber" on my nvme. I have created a PR for this.

The k2 scheduler seems really to be outdated and not worth to test or merge currently.

John-Gee:
Thank you for looking into this!

PR #42, opened 2023-11-11
io-scheduler: Use kyber for nvme's and ssd's

ptr1337:
I have tested kyber for the nvme in various workloads and it does improve the repressives of the desktop under heavy IO scenarios a lot compared to "none". 

Throughput appears to be lower for around 3% in my testings (tested with kdiskmark).

Also see following:

https://github.com/pop-os/default-settings/pull/149

Fixes: https://github.com/CachyOS/CachyOS-Settings/issues/34

### DIFF
--- a/etc/udev/rules.d/60-ioschedulers.rules
+++ b/etc/udev/rules.d/60-ioschedulers.rules
@@ -1,6 +1,7 @@
-# set scheduler for NVMe
-ACTION=="add|change", KERNEL=="nvme[0-9]n[0-9]", ATTR{queue/scheduler}="none"
-# set scheduler for SSD and eMMC
-ACTION=="add|change", KERNEL=="sd[a-z]*|mmcblk[0-9]*", ATTR{queue/rotational}=="0", ATTR{queue/scheduler}="mq-deadline"
-# set scheduler for rotating disks
-ACTION=="add|change", KERNEL=="sd[a-z]*", ATTR{queue/rotational}=="1", ATTR{queue/scheduler}="bfq"
+# BFQ is recommended for slow storage such as rotational block devices and SD cards.
+ACTION=="add|change", SUBSYSTEM=="block", ATTR{queue/rotational}=="1", ATTR{queue/scheduler}="bfq"
+ACTION=="add|change", SUBSYSTEM=="block", KERNEL=="mmcblk?", ATTR{queue/scheduler}="bfq"
+
+# Kyber is recommended for faster storage such as NVME and SATA SSDs.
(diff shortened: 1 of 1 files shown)


John-Gee:
I'm curious what your reasons were for reverting this.

Lately (not sure since when exactly) my system has been crawling under heavy IO and now it makes me wonder if the revert was the cause, but maybe not.

ptr1337:
> I'm curious what your reasons were for reverting this.
> 
> Lately (not sure since when exactly) my system has been crawling under heavy IO and now it makes me wonder if the revert was the cause, but maybe not.
> 
> edit: nope that was not it, it happened again after reverting.

Mainly because we noticed a regression in benchmarks related to that.
Generally the responives should be better with kyber, there you right.

I saw also some regression on heavy IO, it seems like there is some regression in the 6.7 Kernel.

John-Gee:
Oh I am still on 6.6, I'm usually scared of .0 so I have not tried that yet.

If you don't mind me saying, I think it'd be good to note that in the commit and maybe the file too, so that both users like me, but also devs like you in the future, can know why the change was made.

Actually it makes me wonder if you can make the rule depends on the kernel version. I understand people like me are not supported, but this affects LTS users as well.

Thank you!

ptr1337:
> Oh I am still on 6.6, I'm usually scared of .0 so I have not tried that yet.
> 
> If you don't mind me saying, I think it'd be good to note that in the commit and maybe the file too, so that both users like me, but also devs like you in the future, can know why the change was made.
> 
> Actually it makes me wonder if you can make the rule depends on the kernel version. I understand people like me are not supported, but this affects LTS users as well.
> 
> Thank you!

6.7 is pretty okay. Maybe there is some kind of other regression going on? Im not sure.
Sure, I will explain commit messages better and do them via PR.

The regression for nvme's is present through 6.6 also, it is not only 6.7.
We just did recently benchmarks compared to other distros and noted some little regression in the filesystem and stress-ng fork threads, but this seems to be also present on archlinux.
Might be a packaging thing.

John-Gee:
I  appreciate your fast replies and you guys doing benchmarks to keep it all going smoothly!

Thanks a lot!

John-Gee:
With this change: https://git.kernel.dk/cgit/linux/commit/?h=block-deadline&id=e38582207dbf894b18da56981248b1007bd0b675 I'm curious how BFQ will fare in your tests for NVMEs and SSDs.

Although biased, Paolo always seemed to suggest BFQ was it for desktop responsiveness.

PR #49, opened 2024-01-16
Move 'optimize-interruptfreq' from systemd to tmpfiles

rejedai:
The service has been moved because it duplicates the functionality of tmpfiles.


ventureoo:
Mostly I agree with that. But the issue is that pci-latency service may conflict with this. I don't know how relevant this is for now, but should be rechecked.

ventureoo:
Well, nvm, looks like we already have both services enabled.

PR #51, opened 2024-02-08
Use X11 rootless by default for SDDM

ventureoo:
For security purposes and to avoid issues with complete system hangs when missing modules/other situations.

### DIFF
--- a/etc/sddm.conf.d/rootless-x11.conf
+++ b/etc/sddm.conf.d/rootless-x11.conf
@@ -1,0 +1,2 @@
+[General]
+DisplayServer=x11-user


ventureoo:
Marked as Draft, since it completely breaks SDDM for now

1Naim:
Since we now ship SDDM with wayland as default, and going forward we should aim for default wayland in general, I guess this PR is now obsolete?

ventureoo:
> Since we now ship SDDM with wayland as default, and going forward we should aim for default wayland in general, I guess this PR is now obsolete?

Yeah, seems to.

PR #54, opened 2024-03-16
Change I/O scheduling udev rules

damian101:
Finer control over which I/O scheduler is used for a specific storage type.


damian101:
The reason why I set both `ATTR{queue/rotational}=="1", ATTR{queue/write_cache}=="write back"` for HDD was that I had some flash drives that self-reported as rotational. Probably to force selection of a scheduler more friendly for slow storage...
I assume there are practically no DRAM-less HDDs around anymore?

Issue #57, opened 2024-04-10
We should add a .urls file /etc/debuginfod


averms:
Not sure if this is the right repository to report this but currently running `gdb` on my CachyOS installation doesn't download debug infos correctly. The fix is to add a cachyos.urls file in /etc/debuginfod next to the already existing archlinux.urls.

Should be implemented in a way that works for conversions from Arch.

ptr1337:
https://github.com/CachyOS/CachyOS-Settings/pull/58

Could you please test, if it fixes the issue?

ptr1337:
Its now rolled out to users (40-1) and should arrive also to the docker containers soon.

I will make also the weekend some updates to the docker containers and add also a v4 container.

PR #58, opened 2024-04-10
Add CachyOS URL to debuginfod

ptr1337:
Currently debuginfod uses as default the archlinux debuginfod URL, which makes issues when used together with CachyOS. The Archlinux URL is still required, since we dont ship all packages yet.

Fixes: https://github.com/CachyOS/CachyOS-Settings/issues/57

### DIFF
--- a/etc/debuginfod/cachyos.urls
+++ b/etc/debuginfod/cachyos.urls
@@ -1,0 +1,1 @@
+https://debuginfod.cachyos.org


averms:
Yep this works, thanks! Will this also apply to the Docker images?

ptr1337:
> Yep this works, thanks! Will this also apply to the Docker images?

Great!
Yes, we just need to do a new release for cachyos-settings and then it will apply also to the docker containers.
Ill merge this then and will do tomorrow a new release.

PR #62, opened 2024-06-06
feat: add cachy-chroot script

SoulHarsh007:
Initial release of cachy-chroot script. This script simplifies the process of chrooting by prompt-based selection of the block devices and the mount points. It also handles BTRFS subvolumes automatically.


vnepogodin (approved):
LGTM

PR #68, opened 2024-06-23
Remove not used scripts

ptr1337:
This remove not used scripts.

amdpstate-guided is kept, to change from active more easily. active is the default mode.


ventureoo:
We still have le9uo in our kernel-patches. Do we really have to remove the le9uo script?

ptr1337:
> We still have le9uo in our kernel-patches. Do we really have to remove the le9uo script?

Its not directly in linux-cachyos. It was shipped for one minor kernel version, and then droped.

ptr1337:
Also, the script is kinda outdated, since the "new" le9uo should work together with MG-LRU and does not require any disabling.

Also "tunecfs", which is using the old cfs values, which is not used anymore by any of our kernels

ventureoo:
> Also, the script is kinda outdated, since the "new" le9uo should work together with MG-LRU and does not require any disabling.
Okay, then it's all right.

ventureoo (approved):
LGTM

PR #69, opened 2024-06-23
Add helper script to set powerprofiles to performance

ptr1337:
CachyOS does not use Gamemode as default, which makes it a bit harder to switch to the performance governor.

Lets provide a script, which simply utilizes power-profiles-daemon to switch to the performance mode.

User can add to the Game Launch Options:

`game-performance %command% `

todo:

Make it maybe reset, when game does stop

### DIFF
--- a/usr/bin/game-performance
+++ b/usr/bin/game-performance
@@ -1,4 +1,5 @@
 #!/bin/bash
 # Helper script to enable the performance gov with proton or others
 
-powerprofilesctl set performance
+# set performance governors, as long the game is launched
+powerprofilesctl launch -p performance -r "Launched with CachyOS game-performance utility" "$@"
--- a/usr/bin/game-performance
+++ b/usr/bin/game-performance
@@ -1,0 +1,4 @@
+#!/bin/bash
+# Helper script to enable the performance gov with proton or others
+
+powerprofilesctl set performance


vnepogodin:
> User can add to the Game Launch Options:
> game-performance %command%

For that to work you need to run something like [here](https://github.com/CachyOS/kernel-manager/blob/develop/src/rootshell.sh)

vnepogodin:
Also the todo is essential imho

vnepogodin:
> > User can add to the Game Launch Options:
> > game-performance %command%
> 
> For that to work you need to run something like [here](https://github.com/CachyOS/kernel-manager/blob/develop/src/rootshell.sh)

like that:
```bash
#!/bin/bash
# Helper script to enable the performance gov with proton or others

# set perf
powerprofilesctl set performance

# launch what was passed as arguments
exec "$@"
```

usr/bin/game-performance
```diff
@@ -0,0 +1,4 @@
+#!/bin/bash
```
vnepogodin, line 1:
```suggestion
#!/bin/bash
# Helper script to enable the performance gov with proton or others

# set perf and launch what was passed as arguments
powerprofilesctl launch --profile=performance "$@"
```

vnepogodin (changes requested):

PR #71, opened 2024-07-03
Revert "nvidia.conf: Enable GSP again"

Edgars-Cirulis:
Hello there,

The nvidia-dkms 555.58.02-1 still does not completely fix the issues with the GSP firmware. There are still minor stutters on the Plasma desktop. Additionally, enabling V-Sync in Minecraft causes the FPS to drop from a stable 144 FPS to 93 FPS. Without GSP Firmware enabled it remains at 144 FPS with no stuttering.

### DIFF
--- a/usr/lib/modprobe.d/nvidia.conf
+++ b/usr/lib/modprobe.d/nvidia.conf
@@ -24,5 +24,5 @@
 # as the framebuffer is handled by the iGPU. This parameter is marked as
 # experimental, so bugs may occur.
 #
-options nvidia NVreg_UsePageAttributeTable=1 NVreg_InitializeSystemMemoryAllocations=0 NVreg_DynamicPowerManagement=0x02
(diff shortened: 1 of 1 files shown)


ptr1337:
Hi,

Actually I will not merge this, since GSP also provides a bunch of benefits. I would actually add to the Wiki an information, that if the performance is not as expected, that user can add this to a file.

I know, that there are here and there little stutters, but we also want to follow upstream. The non GSP also will not get much tested from NVIDIA anymore, so its better to go with GSP enabled.

Edgars-Cirulis:
> Hi,
> 
> Actually I will not merge this, since GSP also provides a bunch of benefits. I would actually add to the Wiki an information, that if the performance is not as expected, that user can add this to a file.
> 
> I know, that there are here and there little stutters, but we also want to follow upstream. The non GSP also will not get much tested from NVIDIA anymore, so its better to go with GSP enabled.

Alright, copy that. Hopefully, it will be fully fixed someday.

PR #72, opened 2024-07-05
Drop limits

ptr1337:
This is handled by realtime-privileges and rtkit already. Also, ulimit -l reports to be unlimited, which should be not done due security concerns.

### DIFF
--- a/etc/security/limits.d/99-audio.conf
+++ b/etc/security/limits.d/99-audio.conf
@@ -1,2 +1,0 @@
-@audio - rtprio 99
-@audio - memlock unlimited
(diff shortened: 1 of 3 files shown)


ventureoo (approved):
LGTM. Esync will soon be replaced by NTSync, so I don't think it's a issue to remove Esync specific files. But what about `/usr/lib/systemd/user.conf.d/limits.conf`? I don't think there's a security issue with it since it's handled by systemd, but what's the need for it? It's worth checking.

ptr1337:
> LGTM. Esync will soon be replaced by NTSync, so I don't think it's a issue to remove Esync specific files. But what about `/usr/lib/systemd/user.conf.d/limits.conf`? I don't think there's a security issue with it since it's handled by systemd, but what's the need for it? It's worth checking.

Yes, that looks also fine to me, but worth to check.
I merge this for now

PR #75, opened 2024-07-17
game-performance: don't fail if the profile not available

vnepogodin:


usr/bin/game-performance
```diff
@@ -1,5 +1,11 @@
 #!/bin/bash
 # Helper script to enable the performance gov with proton or others
 
+# Don't fail if the cpu driver doesn't support performance power profile
+if ! powerprofilesctl list | grep -q 'performance:'; then
+    exec "$@"
+    exit 0
```
ventureoo, line 7: `exit 0` seems to look redundant here.

vnepogodin: Indeed

ventureoo (approved):

Issue #79, opened 2024-09-03
[RFC] Use systemd-resolved with NetworkManager by default


vnepogodin:
Using systemd-resolved will help with caching of dns requests, making steam downloads faster. While NetworkManager still manage other network stack.

Additionally fixes compatibility with some VPN apps which want to use resolved

requires `systemd-resolved.service` to be enabled

ref: https://wiki.archlinux.org/title/NetworkManager#systemd-resolved

1Naim:
There were discussions a while back to consider [dnsproxy](https://github.com/AdguardTeam/dnsproxy) instead. Did it fall short in terms of features provided, e.g. ECH not supported, or do you think its better to use something thats already installed as default?

vnepogodin:
> There were discussions a while back to consider [dnsproxy](https://github.com/AdguardTeam/dnsproxy) instead. Did it fall short in terms of features provided, e.g. ECH not supported, or do you think its better to use something thats already installed as default?

Yeah there were also discussions about not using networkmanager at all, in favor of systems stack. Generally some software expects resolved for setting system DNS(like VPN) and fails its not present. Using resolved also doesnt require us to install additional package. For using ECH and DoH, we can just conditionally install it by Welcome app

PR #80, opened 2024-09-03
networkmanager: use systemd-resolved as default dns resolver

vnepogodin:
implements config changes required for https://github.com/CachyOS/CachyOS-Settings/issues/79

### DIFF
--- a/usr/lib/NetworkManager/conf.d/dns.conf
+++ b/usr/lib/NetworkManager/conf.d/dns.conf
@@ -1,0 +1,2 @@
+[main]
+dns=systemd-resolved


1Naim:
What if a user is already using a different dns resolver, e.g. aformentioned dnsproxy in #79 or cloudflare-warp?

vnepogodin:
> What if a user is already using a different dns resolver, e.g. aformentioned dnsproxy in #79 or cloudflare-warp?

If the user already configured networkmanager to use dnsproxy in /etc, then our config will be overwritten. dnsproxy will be still used instead of systemd-resolved

1Naim (approved):
Seems to work with cloudflare-warp. It adds these two extra lines in `/etc/resolv.conf` on top of what cloudflare-warp sets

```bash
# /etc/resolv.conf

# This file was generated by cloudflare-warp.
nameserver 127.0.2.2
nameserver 127.0.2.3
options edns0
options trust-ad

ptr1337 (approved):

iamshadow2008:
Systemd-resolved breaks my personal setup. I have my own caching DNS server set up on my network, using dnscrypt-proxy and NextDNS. After installing the cachyos-settings update that includes this change, my two systems can no longer resolve any hosts. Disabling systemd-resolved and removing the usr/lib/NetworkManager/conf.d/dns.conf file and rebooting, systems work fine again. Will I have to do that every time the package updates?

Even if it did "work" I would prefer to not have multiple large independent caches as that could cause problems.

vnepogodin:
> Systemd-resolved breaks my personal setup. I have my own caching DNS server set up on my network, using dnscrypt-proxy and NextDNS (separate device hosting this and other services). After installing the cachyos-settings update that includes this change, my two systems can no longer resolve any hosts. Disabling systemd-resolved and removing the usr/lib/NetworkManager/conf.d/dns.conf file and rebooting, systems work fine again. Will I have to do that every time the package updates?
> 
> Even if it did "work" I would prefer to not have multiple large independent caches as that could cause problems.

well if you had already setup the networkmanager to use custom dns, e.g dnscrypt-proxy or dnsmasq then the settings update shouldnt cause any changes on your side.

if you put (for example) to `/etc/NetworkManager/conf.d/dns.conf` then resolved wont be used as our settings gets overwritten with config in /etc

https://wiki.archlinux.org/title/NetworkManager#dnsmasq
https://wiki.archlinux.org/title/NetworkManager#Unmanaged_/etc/resolv.conf

iamshadow2008:
No client side configuration was done, as my dns server is network-wide. When connecting to the network the dns server information is sent to the client thru dhcp.

TheWyn:
> if you had already setup the networkmanager to use custom dns, e.g dnscrypt-proxy or dnsmasq then the settings update shouldnt cause any changes on your side.

With systemd-resolved now enabled by default should dnsmasq not be installed by default in calamares?

PR #81, opened 2024-09-06
sbctl-batch-sign: Add workaround for /boot and /boot/efi partitioned systems

1Naim:
When using GRUB or rEFInd, sometimes the kernel images that live in `/boot` does not get detected by `sbctl` because `sbctl` is only scanning for the EFI partition `/boot/efi`. This may be a limitation in sbctl itself or something with how we partition our installs and needs to be investigated further. 

- [x] Draft because I need to remove duplicate entries but it should be an adequate solution, just not elegant.

### DIFF
--- a/usr/bin/sbctl-batch-sign
+++ b/usr/bin/sbctl-batch-sign
@@ -10,7 +10,7 @@ if [ "$(id -u)" -ne 0 ]; then
 fi
 
 if [ "$#" -eq 0 ]; then
-    for entries in $(sbctl verify | grep signed | cut -d' ' -f2); do
+    for entries in $(sort -u -i <(sbctl verify | grep 'signed' | cut -d' ' -f2) -i <(find /boot -maxdepth 1 -type f | grep vmlinuz)); do
         sbctl sign -s $entries
     done
 fi


usr/bin/sbctl-batch-sign
```diff
@@ -10,7 +10,10 @@ if [ "$(id -u)" -ne 0 ]; then
 fi
 
 if [ "$#" -eq 0 ]; then
-    for entries in $(sbctl verify | grep signed | cut -d' ' -f2); do
+    sbctl verify | grep 'not signed' | cut -d' ' -f2 > /tmp/entries
+    find /boot -maxdepth 1 -type f | grep vmlinuz >> /tmp/entries
```
1Naim, line 14: So this is the only way I know how to do this. Please let me know if there's a better way but as for now I am not aware.

This will always list all kernel images that are in `/boot`, regardless of whatever it has already been signed or not because `find` cannot discern between a signed image or not.

```bash
❯ sudo ./sbctl-batch-sign
File has already been signed /boot/vmlinuz-linux-cachyos-lto
File has already been signed /boot/vmlinuz-linux-cachyos-rc-lto
```

If this is the case, maybe it's just fine to leave it as 
```bash
for entries in $(sbctl verify | grep 'not signed' | cut -d' ' -f2 & find /boot -maxdepth 1 -type f | grep vmlinuz); do
```
This will list duplicate entries in the for loop on first time setup, and subsequent runs of the script will always try to sign kernel images, like the proposed solution.

vnepogodin: Duplicate entries can be removed with piping to `uniq`

1Naim: Alright I've did some digging and using process substitution works great. I'm using `sort -u` instead of `uniq` since I couldn't get `uniq` to work so I *think* thats only for txt files. 

```bash
❯ sudo ./sbctl-batch-sign
/boot/EFI/BOOT/BOOTX64.EFI
/boot/EFI/cachyos/fwupdx64.efi
/boot/EFI/systemd/systemd-bootx64.efi
/boot/vmlinuz-linux-cachyos-lto
/boot/vmlinuz-linux-cachyos-rc-lto
/usr/lib/fwupd/efi/fwupdx64.efi
```

vnepogodin: Alright

vnepogodin (approved):
Looks good

PR #87, opened 2024-09-25
timesync: Add config for timesync

ptr1337:
Currently, there are popping several threads in the archlinux reddit up about a broken Time Sync.
Adding the google time sync server seems to solve that. Since we had also issues on the ISO with the proper timesync, lets add a config for this.

As mainserver time.google.com will be used and the other ntp ones as fallback

https://www.reddit.com/r/archlinux/comments/1fp63l8/time_synchronization_not_completing/
https://www.reddit.com/r/archlinux/comments/15zgd4y/comment/k9d2c87/


etc/systemd/timesyncd.conf.d/timesyncd.conf
```diff
@@ -0,0 +1,8 @@
+[Time]
+NTP=time.google.com
+FallbackNTP=0.arch.pool.ntp.org 1.arch.pool.ntp.org 2.arch.pool.ntp.org 3.arch.pool.ntp.org
```
ventureoo, line 3: Maybe make arch ntp servers the primary and google as fallback? Some people may noise about requests to google servers.

ventureoo: Nevermind, looks like Arch's NTP servers are set by default and issue is with them, got it.

ventureoo:
The path should be changed to ``/usr/lib/systemd/timesyncd.conf.d/timesyncd.conf`` so that it is possible to override the config in /etc. Otherwise LGTM.

PR #88, opened 2024-09-28
Improvements on audio, sysctl and nvidia conf

dougg0k:


usr/lib/sysctl.d/99-cachyos-settings.conf
```diff
@@ -33,6 +33,13 @@ vm.dirty_background_bytes = 134217728
 # tunable expresses the interval between those wakeups, in 100'ths of a second (Default is 500).
 vm.dirty_writeback_centisecs = 1500
 
+# Sets the minimum amount of free memory that the kernel tries to maintain.
+# Ensures a buffer of free memory to handle sudden memory demands, which helps in maintaining system stability.
+vm.min_free_kbytes=1000000
```
ptr1337, line 38: This setting shouldnt be globally enabled for all users, since its heavily dependant on the configuration and setup

usr/lib/sysctl.d/99-cachyos-settings.conf
```diff
@@ -49,10 +56,19 @@ kernel.kptr_restrict = 2
 # Disable Kexec, which allows replacing the current running kernel.
 kernel.kexec_load_disabled = 1
 
+# Configures the system to reboot automatically 10 seconds after a kernel panic, assisting in recovery from critical errors.
+kernel.panic = 10
```
ptr1337, line 60: Why? The user has no possibility then to get info about the crash or equal

usr/lib/sysctl.d/99-cachyos-settings.conf
```diff
@@ -49,10 +56,19 @@ kernel.kptr_restrict = 2
...
+
+# Disable magic keys
+kernel.sysrq = 0
```
ptr1337, line 63: Why?

usr/lib/sysctl.d/99-cachyos-settings.conf
```diff
@@ -49,10 +56,19 @@ kernel.kptr_restrict = 2
...
+
 # Increase the maximum connections
 # The upper limit on how many connections the kernel will accept (default 4096 since kernel version 5.6):
 net.core.somaxconn = 8192
 
+# Set the more modern option, even though CPU usage are a bit higher.
+net.core.default_qdisc=cake
```
ptr1337, line 70: Cake can be super problematic on some loads.

usr/lib/sysctl.d/99-cachyos-settings.conf
```diff
@@ -85,15 +101,51 @@ net.ipv4.tcp_slow_start_after_idle = 0
...
+# This helps defend against smurf attacks by ignoring ICMP echo requests to broadcast addresses.
+net.ipv4.icmp_echo_ignore_broadcasts = 1
+
+# Disable IP Forwarding
+# This prevents the system from forwarding packets, crucial for non-router devices for security reasons.
+net.ipv4.ip_forward=0
+
+# Enables low-latency TCP mode.
+# Reduces delay in packet delivery, beneficial for real-time applications.
+net.ipv4.tcp_low_latency = 1
```
ptr1337, line 118: "for real-time applications" - which falling under this?

I think this will also reduce the throughput, should be benched with iperf in loca network

ptr1337 (commented):
Added some comments, about the Audio Stuff im really unsure, if this should be applied.

We may ask a bunch of user for testing, if this causes problems. Im not deeply into this.

@ventureoo Can you review the nvidia stuff?

dougg0k:
Feel free to remove anything you might not want.

usr/lib/sysctl.d/99-cachyos-settings.conf
dougg0k: It's mostly intended to give preference to lower latency.
https://www.ibm.com/docs/en/linux-on-systems?topic=tuning-tcpip-ipv4-settings#wkvm_tune_tcpip__title__5

usr/lib/sysctl.d/99-cachyos-settings.conf
dougg0k: Can be removed.

usr/lib/modprobe.d/disable-mouse-pooling.conf
```diff
@@ -0,0 +1,4 @@
+# https://github.com/kbakdev/mouse-fix/blob/main/mousefix.sh
+# https://wiki.archlinux.org/title/Mouse_polling_rate
```
ptr1337, line 2: Is this even a thing anymore?
Like, normally most mouse should work automatically on 1000Hz

usr/lib/modprobe.d/disable_intel_audio_powersave.conf
```diff
@@ -0,0 +1 @@
+options snd_hda_intel power_save=0
```
ptr1337, line 1: Why? Any references?

usr/lib/sysctl.d/99-cachyos-settings.conf
dougg0k: I would say reduce a vulnerability point, but that would probably be just lack of better understading.

usr/lib/sysctl.d/99-cachyos-settings.conf
dougg0k: That would have intended against frozen systems.

usr/lib/sysctl.d/99-cachyos-settings.conf
dougg0k: Ok.

ptr1337:
Actually, generally we do not want to provide any "improvements", which can be found somehwere in 2014, where linux was in a different state. We really need to carefully review these and find references, as well as need to retest if this is **really** an improvement.

Often, such tweaks can also introduce issue, therefore should be tested.

The sound changes are generally looking good, but we need to test this a bit more. 

NVIDIA:

I have checked related to rebar, and CPU-X is reporting that rebar is enabled. Checking "sudo dmesg | grep BAR" does not give me any info, how much is used. Will retest with the variable.

On the other NVIDIA stuff, we need to check, if this can be really globally applied and brings an improvement.

usr/lib/modprobe.d/disable_intel_audio_powersave.conf
dougg0k: Default

```
nu ❯ cat /sys/module/snd_hda_intel/parameters/power_save | to text
10
```

usr/share/pipewire/pipewire.conf.d/10-clock-rate.conf
```diff
@@ -0,0 +1,7 @@
+# Help reduce / avoid pop / crackling
+context.properties = {
+	default.clock.allowed-rates = [ 44100 48000 96000 192000 ]
+	default.clock.quantum = 2048
+  default.clock.min-quantum = 32
+  default.clock.max-quantum = 8192
```
1Naim, line 6: Why? This it not mitigating any crackling. Usually crackling happens when applications choose to use a quant of 240, and since pipewire defaults to 2^n for output quant, it will use the 128. The default values of min = 32, default = 1024 and max = 2048 is already fine. If you want to mitigate crackling, bump min quant, but that is also not ideal.

usr/lib/modprobe.d/nvidia.conf
```diff
@@ -4,32 +4,45 @@
...
 # NVreg_EnableGpuFirmware=0 (Default 1) - Disables GSP Firmware on NVIDIA's
 # closed source kernel modules. This option is ignored when using NVIDIA's
 # open kernel modules.
-#
+
+# Disable GSP Firmware on NVIDIA's closed source kernel modules
+# options nvidia NVreg_EnableGpuFirmware=0
+
+# NVreg_EnableResizableBar=1 (Default 0) - Enable Resizable Bar support.
+options nvidia NVreg_EnableResizableBar=1
```
ventureoo, line 30: The reason I didn't want to enable before was that BAR1 seemed to work for me without the parameter on my laptop. I think it might make sense for desktop video cards, but I'm not sure about potential issues like that: https://www.reddit.com/r/linux_gaming/comments/119hwmt/this_setting_may_help_vkd3d_games_that_have/

usr/lib/sysctl.d/99-cachyos-settings.conf
```diff
@@ -85,15 +101,51 @@ net.ipv4.tcp_slow_start_after_idle = 0
...
+# Protects from attackers that are using ip spoofing methods to do harm
+net.ipv4.conf.all.rp_filter = 1
+
+# Prevent ICMP Broadcast Echo Requests
+# This helps defend against smurf attacks by ignoring ICMP echo requests to broadcast addresses.
+net.ipv4.icmp_echo_ignore_broadcasts = 1
+
+# Disable IP Forwarding
+# This prevents the system from forwarding packets, crucial for non-router devices for security reasons.
+net.ipv4.ip_forward=0
```
ventureoo, line 114: It's already disabled by default.

usr/lib/modprobe.d/disable_intel_audio_powersave.conf
ptr1337: Yes, but how much more power it will use?
Also, does disabling powersave give there an improvement?

dougg0k:
On the nvidia, I checked through here `cat /proc/driver/nvidia/params`, though the values were only changed after a `sudo mkinitcpio -P`

usr/share/pipewire/pipewire.conf.d/10-clock-rate.conf
```diff
@@ -0,0 +1,7 @@
+# Help reduce / avoid pop / crackling
+context.properties = {
+	default.clock.allowed-rates = [ 44100 48000 96000 192000 ]
+	default.clock.quantum = 2048
```
ventureoo, line 4: All quantum parameters are strictly dependent on the sampling rate used, which is device dependent. PipeWire already counts these values optimally by default. The only thing I have left to do to combat audio quacking is to increase min-quantum, as the Gentoo Wiki advises to do.
https://wiki.gentoo.org/wiki/PipeWire/en#Crackling_and_stuttering

usr/share/wireplumber/wireplumber.conf.d/50-alsa-config.conf
```diff
@@ -0,0 +1,16 @@
+# https://wiki.archlinux.org/title/PipeWire#Audio_cutting_out_when_multiple_streams_start_playing
```
1Naim, line 1: This should not be globally enabled for users.

usr/share/wireplumber/wireplumber.conf.d/51-disable-suspension.conf
```diff
@@ -0,0 +1,35 @@
+# https://wiki.archlinux.org/title/PipeWire#Noticeable_audio_delay_or_audible_pop/crack_when_starting_playback
```
1Naim, line 1: Same as above.

usr/share/wireplumber/wireplumber.conf.d/51-disable-suspension.conf
```diff
@@ -0,0 +1,35 @@
...
+      {
+        node.name = "~alsa_input.*"
+      },
+      {
+        node.name = "~alsa_output.*"
+      }
+    ]
+    actions = {
+      update-props = {
+        session.suspend-timeout-seconds = 0
```
ventureoo, line 14: Same as in the case of snd-hda-intel.

usr/lib/modprobe.d/nvidia.conf
```diff
@@ -4,32 +4,45 @@
 # specific address mapped inside the register and utilizes the memory
 # architecture and instruction set more efficiently and faster. If your system
 # can support this feature, it should improve CPU performance.
-#
+options nvidia NVreg_UsePageAttributeTable=1
```
1Naim, line 7: Whats the point of splitting each modprobe option?

usr/lib/modprobe.d/nvidia.conf
```diff
@@ -4,32 +4,45 @@
...
+
 # nvidia_drm.modeset=1 (default 0) - Enables modesetting support for the NVIDIA
 # driver. Critical for Wayland support and proper PRIME Offload operation.
 #
 # nvidia_drm.fbdev=1 - Enables hardware framebuffer support. Allows to use
 # native display resolution in tty. This option has no effect on PRIME laptops,
 # as the framebuffer is handled by the iGPU. This parameter is marked as
 # experimental, so bugs may occur.
-#
-options nvidia NVreg_UsePageAttributeTable=1 NVreg_InitializeSystemMemoryAllocations=0 NVreg_DynamicPowerManagement=0x02
```
ventureoo, line 31: I'd like to keep options enabling to a single line.

usr/lib/modprobe.d/disable-mouse-pooling.conf
dougg0k: Maybe I mistunderstood how it works, but the idea is to have something, so the system would not be the sort of bottleneck on the matter.

usr/lib/modprobe.d/nvidia.conf
dougg0k: Keep below the description. Does it matter?

usr/share/wireplumber/wireplumber.conf.d/51-disable-suspension.conf
dougg0k: Shouldnt this help?

usr/lib/modprobe.d/nvidia.conf
```diff
@@ -4,32 +4,45 @@
...
+
+# Disable GSP Firmware on NVIDIA's closed source kernel modules
+# options nvidia NVreg_EnableGpuFirmware=0
+
+# NVreg_EnableResizableBar=1 (Default 0) - Enable Resizable Bar support.
+options nvidia NVreg_EnableResizableBar=1
+
+# Vreg_EnableStreamMemOPs=1 (Default 0) - Activates the support for
+# CUDA Stream Memory Operations in user-mode applications.
+options nvidia NVreg_EnableStreamMemOPs=1
```
ventureoo, line 34: Too highly specific and rarely used parameter. But overall I'm fine with it.

usr/share/wireplumber/wireplumber.conf.d/50-alsa-config.conf
dougg0k: Sure

usr/share/pipewire/pipewire.conf.d/10-clock-rate.conf
dougg0k: Ok

usr/share/pipewire/pipewire.conf.d/10-clock-rate.conf
```diff
@@ -0,0 +1,7 @@
+# Help reduce / avoid pop / crackling
+context.properties = {
+	default.clock.allowed-rates = [ 44100 48000 96000 192000 ]
```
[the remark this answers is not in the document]
dougg0k: Right.

dougg0k:
It seems nothing is wanted, you can just close the PR.

dougg0k:
If you want anything specifically, feel free to use and add yourself.

1Naim:
We didn't want to discourage you from contributing, it's just you should have tested each and every option and understood a bit deeply into how each config setting works and how it weighs between the pros and cons. I've seen that you've cited some sources for the config options but you didn't go to detail as to: 1. why it should be added, is it a problem that affects the majority of the userbase 2. Does the benefits of setting these config options outweight any possible downside from setting it

ventureoo:
> If you want anything specifically, feel free to use and add yourself.

I'm sorry, but that's just too much for one PR. You can split it and then we can test this properly and discuss what changes need to be made. Any settings should be something that works well for everyone, edge cases are something that should be described in the wiki or done under some condition.

usr/share/pipewire/pipewire.conf.d/10-clock-rate.conf
dougg0k: Well, I had crackling issues, after adding these all together, it was gone. So, might have been one or the other that fixed.

usr/share/wireplumber/wireplumber.conf.d/51-disable-suspension.conf
dougg0k: Same as above, usually, anything power save, I would disable, but that might not be optimal to everyone.

usr/lib/modprobe.d/nvidia.conf
dougg0k: Pretty sure I saw this file on another repo. Not sure of the one line benefits.

usr/lib/modprobe.d/disable_intel_audio_powersave.conf
dougg0k: No idea. My idea were always, disable anything power save related in attempt to reduce latency and improve responsiveness, I dont always measure, nor care to.

usr/lib/modprobe.d/nvidia.conf
dougg0k: Just like resizable bar and any other, it should be there if available, otherwise, no harm?

usr/lib/modprobe.d/nvidia.conf
```diff
@@ -4,32 +4,45 @@
...
+
+# Enable Modern Standby, known to be faster on suspend / resume.
+# https://download.nvidia.com/XFree86/Linux-x86_64/560.35.03/README/powermanagement.html
+options nvidia NVreg_EnableS0ixPowerManagement=1
```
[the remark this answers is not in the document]
dougg0k: If both the platform and the NVIDIA GPU support S0ix-based power management, then the NVIDIA Linux driver will put the GPU video memory in self refresh mode during s2idle system suspend. S0ix-based suspend will consume more power than legacy S3 system suspend, but it will enter and exit suspend/resume more quickly. Also, the suspend/resume latency will be constant irrespective of GPU video memory usage.

usr/lib/modprobe.d/nvidia.conf
ventureoo: It's not about benefits, it's just that it's common practice to specify multiple parameters for a single module through a space, rather than breaking up individual lines. Comments in any case name the parameter they refer to.

dougg0k: It's easier to read if each with their own description.

dougg0k: It's easier to read if each with their own description. And why it's a common practice again? It doesnt add to anything. Seems like a want only. Anyhow, it seems that it will stay that way.

usr/lib/modprobe.d/nvidia.conf
dougg0k: Only after adding through this command, it appeared enabled on `/proc/driver/nvidia/params`

ventureoo: > Only after enabling through this command, it appeared enabled on `/proc/driver/nvidia/params`

My point was that whether this parameter is enabled or not I still get the same BAR1 memory values via ``nvidia-smi dmon -s m``.

PR #90, opened 2024-09-29
nvidia-application-profiles: Add profile to avoid aggressive vram usage

ptr1337:
Currently the VRAM buffer can be pretty fast bloated, due a bug in the driver.

Avoid this issue with setting an application profile. This has been suggested by a NVIDIA Developer, see below:

https://github.com/NVIDIA/egl-wayland/issues/126#issuecomment-2379945259


ventureoo:
Well, I originally assumed that we should put this in nvidia-utils

ptr1337:
> Well, I originally assumed that we should put this in nvidia-utils

Right, will do

PR #92, opened 2024-09-30
Audio quality and power states

dougg0k:
There is a setting to improve audio quality and the rest are to disable suspend / power save states.

Disabling those, can help avoid pop, crackling, and any sort of distortions.

### DIFF
--- a/usr/share/pipewire/pipewire.conf.d/15-resample.conf
+++ b/usr/share/pipewire/pipewire.conf.d/15-resample.conf
@@ -1,0 +1,3 @@
+stream.properties = {
+  resample.quality = 10
+}
--- a/usr/lib/modprobe.d/audio.conf
+++ b/usr/lib/modprobe.d/audio.conf
@@ -1,0 +1,2 @@
+options snd_hda_intel power_save=0
+options snd_hda_intel power_save_controller=N
--- a/usr/share/wireplumber/wireplumber.conf.d/51-disable-suspension.conf
+++ b/usr/share/wireplumber/wireplumber.conf.d/51-disable-suspension.conf
@@ -1,0 +1,34 @@
+monitor.alsa.rules = [
+  {
+    matches = [
+      {
+        node.name = "~alsa_input.*"
+      },
+      {
+        node.name = "~alsa_output.*"
+      }
+    ]
+    actions = {
+      update-props = {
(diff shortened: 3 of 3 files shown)


1Naim:
As mentioned previously, we try to avoid touching power saving features as much as possible. We usually let dedicated programs handle those (e.g. power-profiles-daemon). 

> Disabling those, can help avoid pop, crackling, and any sort of distortions

Have you measured the increase in power consumption by disabling these settings?

usr/share/pipewire/pipewire.conf.d/15-resample.conf
```diff
@@ -0,0 +1,3 @@
+stream.properties = {
+  resample.quality = 10
```
1Naim, line 2: This seems fine to me tho.

dougg0k:
No, I did not, nor would care to. Just like audio quality, the continous better experience comes before anything like that.

usr/share/pipewire/pipewire.conf.d/15-resample.conf
monarc99: It just doesn't make much sense as a default. For example, you wouldn't hear any improvement on a laptop anyway and it would only use up battery power unnecessarily. 

Rather something for the wiki, if it brings something.

dougg0k:
Feel free to close the PR if nothing is wanted.

usr/share/pipewire/pipewire.conf.d/15-resample.conf
1Naim: That's very valid. For one thing, this is only used when the audio is routed via pipewire-pulse. Secondly, it won't be used at all when a user has allowed more sample rates than the default (48000).

1Naim:
Yeah I think I'm going to close this. It seems a bit low effort and the problems this is trying to fix are very niche, i.e. not a lot of people experience it.

ventureoo:
I've already suggested an approach where we could switch to using tuned and tuned-pdd to differentiate between tuning for battery-powered and DC-powered devices. Unfortunately, we have stopped using power-profiles-daemon for now, but we can still use the approach where the initial values are restored using udev rules, when device is going to use battery.

dougg0k:
https://fedoraproject.org/wiki/Changes/TunedReplacesPower-profiles-daemon

PR #93, opened 2024-09-30
udev additional rules

dougg0k:
There are two rules added

1- To help avoid freezes in the system on nvmes.
2- Disable usb auto suspend to help prevent problems with usb connections.


ventureoo:
>  2- Disable usb auto suspend to help prevent problems with usb connections.
What kind of problems you are talking about?

dougg0k:
Problems of itself, auto disconecting devices. Creating delay / lag due to entering suspended mode.

dougg0k:
This is not moving forward, if you decided to add, just merge, after.

PR #94, opened 2024-09-30
black / freeze screen on gnome fix

dougg0k:
A solution that fixed a inactivity problem here and it seems to do more than that.

https://discuss.cachyos.org/t/second-monitor-black-screen/3129

https://github.com/robswc/ubuntu-22-nvidia-suspend-fix-script

If anything, should be optional. 

```
systemctl enable gnome-shell-suspend
systemctl enable gnome-shell-resume
```


dougg0k:
This is not moving forward, if you decided to add, just merge, after.

PR #102, opened 2024-10-04
sysctl additionals

dougg0k:
https://github.com/CryoByte33/steam-deck-utilities/blob/main/docs/tweak-explanation.md#swappiness

### DIFF
--- a/usr/lib/sysctl.d/99-cachyos-settings.conf
+++ b/usr/lib/sysctl.d/99-cachyos-settings.conf
@@ -1,7 +1,7 @@
 # The sysctl swappiness parameter determines the kernel's preference for pushing anonymous pages or page cache to disk in memory-starved situations.
 # A low value causes the kernel to prefer freeing up open files (page cache), a high value causes the kernel to try to use swap space,
 # and a value of 100 means IO cost is assumed to be equal.
-vm.swappiness = 100
+vm.swappiness = 1
 
 # The value controls the tendency of the kernel to reclaim the memory which is used for caching of directory and inode objects (VFS cache).
 # Lowering it from the default value of 100 makes the kernel less inclined to reclaim VFS cache (do not set it to 0, this may produce out-of-memory conditions)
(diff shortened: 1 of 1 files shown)


usr/lib/sysctl.d/99-cachyos-settings.conf
```diff
@@ -74,6 +77,30 @@ net.ipv4.tcp_ecn = 1
...
+# Avoids fragmentation and improves performance by discovering the optimal MTU size.
+# https://xn--tigreray-i1a.org/english/post/15-the-strange-case-of-the-tcp-black-holes/
+net.ipv4.tcp_mtu_probing = 1
+
+# Controls the handling of ICMP redirect messages and source-routed packets.
+# Enhances security by disabling these features, preventing potential man-in-the-middle attacks.
+# (we are not a router)
+net.ipv4.conf.all.send_redirects = 0
+net.ipv4.conf.all.accept_redirects = 0
+net.ipv4.conf.all.accept_source_route = 0
```
1Naim, line 102: This is already the default

usr/lib/sysctl.d/99-cachyos-settings.conf
```diff
@@ -33,6 +33,9 @@ vm.dirty_background_bytes = 134217728
 # tunable expresses the interval between those wakeups, in 100'ths of a second (Default is 500).
 vm.dirty_writeback_centisecs = 1500
 
+# https://serverfault.com/a/1142151
+vm.overcommit_memory = 1
```
1Naim, line 37: Can you tell me why setting this better than the default? The link you cited and the kernel documentation also mentions that the default value of 0 prevents BS allocations and imo is what it should be at.

usr/lib/sysctl.d/99-cachyos-settings.conf
```diff
@@ -1,7 +1,7 @@
 # The sysctl swappiness parameter determines the kernel's preference for pushing anonymous pages or page cache to disk in memory-starved situations.
 # A low value causes the kernel to prefer freeing up open files (page cache), a high value causes the kernel to try to use swap space,
 # and a value of 100 means IO cost is assumed to be equal.
-vm.swappiness = 100
+vm.swappiness = 1
```
1Naim, line 4: No. We want swappiness to be high because we are using ZRAM. Even if we weren't using ZRAM, having a swappiness value of 1 is still very broken.

PR #105, opened 2024-10-04
game-performance: Fix incorrect argument passing

ventureoo:
Besides, specifying `-p` seems redundant. This is the default performance.


vnepogodin (approved):
yeah that's fair given that steam wrapper mangles arguments

ptr1337 (approved):

PR #107, opened 2024-10-24
Various sysctl cleanups

1Naim:
Currently there seems to be a bit of a redundancy in the settings set here, so I'd like to clean it up a bit. There are also some removals/tweaks to settings that don't make quite sense to me (based on my limited knowledge), so I've opted to removed them in the PR, but I am very open for discussion.

### DIFF
--- a/usr/lib/sysctl.d/99-cachyos-settings.conf
+++ b/usr/lib/sysctl.d/99-cachyos-settings.conf
@@ -85,9 +85,6 @@ net.ipv4.tcp_slow_start_after_idle = 0
 # Protect against tcp time-wait assassination hazards, drop RST packets for sockets in the time-wait state. Not widely supported outside of Linux, but conforms to RFC:
 net.ipv4.tcp_rfc1337 = 1
 
-# Set the maximum watches on files
-fs.inotify.max_user_watches = 524288
-
 # Set size of file handles and inode cache
 fs.file-max = 2097152
 
--- a/usr/lib/sysctl.d/99-cachyos-settings.conf
+++ b/usr/lib/sysctl.d/99-cachyos-settings.conf
@@ -5,7 +5,7 @@ vm.swappiness = 100
 
 # The value controls the tendency of the kernel to reclaim the memory which is used for caching of directory and inode objects (VFS cache).
 # Lowering it from the default value of 100 makes the kernel less inclined to reclaim VFS cache (do not set it to 0, this may produce out-of-memory conditions)
-#vm.vfs_cache_pressure=50
+vm.vfs_cache_pressure=50
 
 # Contains, as a bytes of total available memory that contains free pages and reclaimable
 # pages, the number of pages at which a process which is generating disk writes will itself start
--- a/usr/lib/sysctl.d/99-cachyos-settings.conf
+++ b/usr/lib/sysctl.d/99-cachyos-settings.conf
@@ -49,10 +49,6 @@ kernel.kexec_load_disabled = 1
 # Using the value 3 instead of the default 1 allows TCP Fast Open for both incoming and outgoing connections:
 net.ipv4.tcp_fastopen = 3
 
-# TCP SYN cookie protection
-# Helps protect against SYN flood attacks. Only kicks in when net.ipv4.tcp_max_syn_backlog is reached:
-net.ipv4.tcp_syncookies = 1
-
 # TCP Enable ECN Negotiation by default
 net.ipv4.tcp_ecn = 1
 
--- a/usr/lib/sysctl.d/99-cachyos-settings.conf
+++ b/usr/lib/sysctl.d/99-cachyos-settings.conf
@@ -49,10 +49,6 @@ kernel.kptr_restrict = 2
 # Disable Kexec, which allows replacing the current running kernel.
 kernel.kexec_load_disabled = 1
 
-# Increase the maximum connections
-# The upper limit on how many connections the kernel will accept (default 4096 since kernel version 5.6):
-net.core.somaxconn = 8192
-
 # Enable TCP Fast Open
 # TCP Fast Open is an extension to the transmission control protocol (TCP) that helps reduce network latency
 # by enabling data to be exchanged during the sender’s initial TCP SYN [3].
--- a/usr/lib/sysctl.d/99-cachyos-settings.conf
+++ b/usr/lib/sysctl.d/99-cachyos-settings.conf
@@ -23,12 +23,6 @@ vm.page-cluster = 0
 # dirty data.
 vm.dirty_background_bytes = 134217728
 
-# This tunable is used to define when dirty data is old enough to be eligible for writeout by the
-# kernel flusher threads.  It is expressed in 100'ths of a second.  Data which has been dirty
-# in-memory for longer than this interval will be written out next time a flusher thread wakes up
-# (Default is 3000).
-#vm.dirty_expire_centisecs = 3000
-
 # The kernel flusher threads will periodically wake up and write old data out to disk.  This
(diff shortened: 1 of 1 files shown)


1Naim (commented):
Guess I can't add further snippet reviews so I'll add it here

```
# This tunable is used to define when dirty data is old enough to be eligible for writeout by the
# kernel flusher threads.  It is expressed in 100'ths of a second.  Data which has been dirty
# in-memory for longer than this interval will be written out next time a flusher thread wakes up
# (Default is 3000).
#vm.dirty_expire_centisecs = 3000
```
We are also not setting this.

```
# Only experimental!
# Let Realtime tasks run as long they need
# sched: RT throttling activated
kernel.sched_rt_runtime_us=-1
```

I'm skeptical about this, maybe we can set it to a higher predefined value than the default instead?

usr/lib/sysctl.d/99-cachyos-settings.conf
```diff
@@ -49,10 +44,6 @@ kernel.kptr_restrict = 2
 # Disable Kexec, which allows replacing the current running kernel.
 kernel.kexec_load_disabled = 1
 
-# Increase the maximum connections
-# The upper limit on how many connections the kernel will accept (default 4096 since kernel version 5.6):
-net.core.somaxconn = 8192
```
Shendisx, line 54: My opinion is that 8192 is overkill for any user that is not hosting a decently sized server, or idk a database. maybe we could hardcode "8192" for the server kernel variant and leave "4096" which is the current default value in Linux therefore removing or commenting this sysctl.

usr/lib/sysctl.d/99-cachyos-settings.conf
```diff
@@ -1,8 +1,3 @@
-# The sysctl swappiness parameter determines the kernel's preference for pushing anonymous pages or page cache to disk in memory-starved situations.
-# A low value causes the kernel to prefer freeing up open files (page cache), a high value causes the kernel to try to use swap space,
-# and a value of 100 means IO cost is assumed to be equal.
-vm.swappiness = 100
```
ventureoo, line 4: I don't really want to revert it. For me, the preference for pushing out anonymous pages equally as file pages is acceptable on a regular swap, since you almost always use that in tandem with Zswap.

usr/lib/sysctl.d/99-cachyos-settings.conf
1Naim: > maybe we could hardcode "8192" for the server kernel variant

If we want to do this, we need to do something like

```c
#ifdef CONFIG_CACHY
#define <max_connections> 4096
#else
#define <max_connections> 8192
```

We don't have a specific config for the server kernel, it is currently just `CONFIG_CACHY=n`, but even then I think the default 4096 should work fine already for servers.

usr/lib/sysctl.d/99-cachyos-settings.conf
1Naim: I don't really know how systems are handled when they want to use something else other than ZRAM only, i.e. ZRAM + swap or Zswap. For the former, I don't think it matters no?, that we set `vm.swappiness=100` here since the ZRAM udev rule will always take precedent, or does the is the swappiness value of 150 set in the udev rule is specifically only for the ZRAM, and that the swappiness in sysctl is used for other swap devices.

For the latter case, AFAIK Zswap is a different thing than ZRAM (it has similar functionality but different systems (?) so using Zswap would effectively lead to users tampering with the udev anyway.

Correct me if I'm wrong on the above but that's how I'm understanding things at this moment.

ventureoo: > I don't really know how systems are handled when they want to use something else other than ZRAM only, i.e. ZRAM + swap or Zswap. For the former, I don't think it matters no?, that we set `vm.swappiness=100` here since the ZRAM udev rule will always take precedent, or does the is the swappiness value of 150 set in the udev rule is specifically only for the ZRAM, and that the swappiness in sysctl is used for other swap devices.
> 

The original purpose of my PR (https://github.com/CachyOS/CachyOS-Settings/pull/19) was to improve swap handling in general. It was not aimed only at ZRAM configurations.

> For the latter case, AFAIK Zswap is a different thing than ZRAM (it has similar functionality but different systems (?) so using Zswap would effectively lead to users tampering with the udev anyway.

Zswap is literally just a buffer between memory and a simple swap that minimizes I/O overhead by temporarily compressing data within the buffer. It doesn't have any block device and there is nothing we can control with udev with respect to Zswap.

usr/lib/sysctl.d/99-cachyos-settings.conf
```diff
@@ -1,8 +1,3 @@
...
-
 # The value controls the tendency of the kernel to reclaim the memory which is used for caching of directory and inode objects (VFS cache).
 # Lowering it from the default value of 100 makes the kernel less inclined to reclaim VFS cache (do not set it to 0, this may produce out-of-memory conditions)
 #vm.vfs_cache_pressure=50
```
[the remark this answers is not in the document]
ventureoo: AFAIK we used to have a patch that increased this value in our kernel. Not sure how things are now.

usr/lib/sysctl.d/99-cachyos-settings.conf
1Naim: Alright, let's keep it then.

usr/lib/sysctl.d/99-cachyos-settings.conf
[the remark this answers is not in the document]
1Naim: https://github.com/CachyOS/linux/commit/9fa4ce33c505a49c63b099a01d63a5438f74d687#diff-d503e0c4ef59449a7b9dd9f14c3f88b35578cdec99edbd05970e0599faf1c324R203 Doesn't seem to be the case now. On another note, we need to rework these tweaks.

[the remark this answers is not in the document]
1Naim: What's the consensus of this? I haven't really read much about this so I'm still not too sure about what impacts a lower value might have compared to the default. Is it worth setting this?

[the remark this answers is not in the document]
ventureoo: Honestly, I'd like to leave it uncommented since we don't have a patch to change it right now, but it's a quite useful tweak, I've been writing more details about it: https://ventureo-codeberg-page.translate.goog/source/kernel-tuning.html?_x_tr_sl=auto&_x_tr_tl=en&_x_tr_hl=en-US&_x_tr_pto=wapp#vfs

1Naim:
This PR seems finished now. @ventureoo when you have the time, it would be nice for this to go through one more review before merging.

usr/lib/sysctl.d/99-cachyos-settings.conf
```diff
@@ -49,20 +43,12 @@ kernel.kptr_restrict = 2
...
 # Using the value 3 instead of the default 1 allows TCP Fast Open for both incoming and outgoing connections:
 net.ipv4.tcp_fastopen = 3
 
-# Enable BBR3
-# The BBR3 congestion control algorithm can help achieve higher bandwidths and lower latencies for internet traffic
-net.ipv4.tcp_congestion_control = bbr
-
 # TCP SYN cookie protection
 # Helps protect against SYN flood attacks. Only kicks in when net.ipv4.tcp_max_syn_backlog is reached:
 net.ipv4.tcp_syncookies = 1
```
ventureoo, line 54:
```suggestion
```
Can be also removed, it's 1 by default:

https://www.kernel.org/doc/Documentation/networking/ip-sysctl.rst

ventureoo (approved):
LGTM

ptr1337 (approved):

PR #108, opened 2024-10-27
tmpfiles.d: Add thp-shrinker config to reduce memory usage

ptr1337:
Using the always mode (which is default in Archlinux and CachyOS) provides a quite big performance improvement compared to madvise, but this comes with an increased memory usage.

Recently META has contributed the THP Shrinker, which allows to split pages which are underutilized. This can reduce the used memory usage, while providing mostly the same performance improvement.

409 is the provide value by META and we might want to search for more suited values on the desktop. Also, we should do some benchmarks (cachyos-benchmarker) to see if there is a massive difference between 509 and 411.

https://lore.kernel.org/lkml/20240830100438.3623486-1-usamaarif642@gmail.com/

### DIFF
--- a/usr/lib/tmpfiles.d/thp-shrinker.conf
+++ b/usr/lib/tmpfiles.d/thp-shrinker.conf
@@ -1,0 +1,6 @@
+# THP Shrinker has been added in the 6.12 Kernel
+# Default Value is 511
+# THP=always policy vastly overprovisions THPs in sparsely accessed memory areas, resulting in excessive memory pressure and premature OOM killing
+# 409 means that any THP that has more than 409 out of 512 (80%) zero filled filled pages will be split.
+# This reduces the memory usage, when THP=always used and the memory usage goes down to around the same usage as when madivise is used, while still providing equal performance improvement
+w! /sys/kernel/mm/transparent_hugepage/khugepaged/max_ptes_none - - - - 409


1Naim:
Why can't we just merge this with [`/usr/lib/tmpfiles.d/thp.conf`](https://github.com/CachyOS/CachyOS-Settings/blob/master/usr/lib/tmpfiles.d/thp.conf), i.e. all in one file. Both are related to Transparent Hugepages and it's not like each file has a lot of content like our sysctl file, so IMO putting this to one file is better.



> Also, we should do some benchmarks (cachyos-benchmarker)

I wonder if this benchmark will show the differences. I guess we need to do a comparison of madvise vs `max_ptes_none=409` vs thp=always default.

1Naim:
kernel_version_comparison_All
Here's some benchmarks with `cachyos-benchmarker`. As it turns out, it seems to be quite unreliable at comparing the differences between `max_ptes_none` values (you can only start to see some regressions with absurd values like 50% of zero filled pages). 

I think a better benchmark for this would be to build a full-LTO kernel, while also checking for memory usage cc @ptr1337 maybe

https://flightlesssomething.ambrosia.one/benchmark/651 Here's a gaming benchmark I did with `stress-ng` in the background as well with some virtual memory flags but there's no difference here either.

ptr1337:
I will merge this for now. It seems not to increase regressions or equal. We can reevaluate the value in the future.

I want to keep it in an own config, so the different functions are split each

PR #118, opened 2024-12-14
Adding optimal performance values

Terromur:
I may have broken English, but will try to explain for each parameter, but you can read my comments in these commits. 

This values specifies the maximum length (bytes) of messages sent from one process to another. Inter-process messaging is done in the kernel's memory and is not swapped to disk, so increasing the value increases the amount of memory used by the operating system: [Link](https://docs.oracle.com/cd/E18887_01/html/e18878/ins_asap_preinstall.htm)
`kernel.msgmax = 65535`
`kernel.msgmnb = 65535`

Increase the TCP SYN queue length so that the system can handle more concurrent connections. Allows to create a large queue to help handle bursts of incoming connection requests or scenarios with high connection speeds. Values that are too high may increase memory consumption.
`net.ipv4.tcp_max_syn_backlog = 8192`

The tcp_keepalive_intvl determines the interval between keepalive packets if no response is received from the remote host. Setting this value too low can lead to network congestion, while setting it too high may slow down the detection of dead connections. A value between 30 and 120 seconds is usually sufficient.
`net.ipv4.tcp_keepalive_intvl = 60`

You can read about the values here: [Link](https://blog.cloudflare.com/optimizing-tcp-for-high-throughput-and-low-latency/)
`net.ipv4.tcp_rmem = 8192 262144 536870912`
`net.ipv4.tcp_wmem = 4096 16384 536870912`
`net.ipv4.tcp_adv_win_scale = -2`
`net.ipv4.tcp_collapse_max_bytes = 6291456`
`net.ipv4.tcp_notsent_lowat = 131072`

Value that specifies the maximum size of the queue of waiting connections. It is recommended to change the parameter either equal to the `backlog` or 2 times the `backlog`.
`net.core.somaxconn = 16384`

Value defines the maximum number of sockets in the TIME_WAIT state that can exist simultaneously. This parameter affects network performance, especially in highly loaded systems Also for this parameter can meet the value 2000000 here [Link](https://wiki.archlinux.org/title/Sysctl)

But it is extremely high, may overload memory, but may partially help against simple DDOS Attacks. (I would leave the optimal value I prescribed). Especially when we've already changed the value to a higher value that increased security. Default value: 262144
`net.ipv4.tcp_max_tw_buckets = 1440000`

Open the TIME-WAIT socket reuse feature, which is very effective for Web servers and Desktops that have a large number of connections.  This helps to avoid running out of available network sockets
`net.ipv4.tcp_tw_reuse = 1`

thresh1 - the minimum number of entries that should be in the ARP cache. If the number of entries is less than this value, the garbage collector will not clear the ARP cache
thresh2 - soft limit on the number of entries in the ARP cache. If the number of entries reaches this value, the garbage collector will start within 5 seconds.
thresh3 - a hard limit on the number of entries in the ARP cache. If the number of entries reaches this value, the garbage collector will start immediately.
`net.ipv4.neigh.default.gc_thresh1 = 1024`
`net.ipv4.neigh.default.gc_thresh2 = 2048`
`net.ipv4.neigh.default.gc_thresh3 = 4096`

Increases the maximum queue length for ARP proxy entries.
`net.ipv4.neigh.default.proxy_qlen = 96`

Sets the maximum number of ARP requests. Values that are too high are excessive and require additional memory and CPU time
`net.ipv4.neigh.default.unres_qlen = 6`

### DIFF
--- a/usr/lib/sysctl.d/99-cachyos-settings.conf
+++ b/usr/lib/sysctl.d/99-cachyos-settings.conf
@@ -78,8 +78,6 @@ net.ipv4.neigh.default.gc_thresh1 = 32
 net.ipv4.neigh.default.gc_thresh2 = 1024
 net.ipv4.neigh.default.gc_thresh3 = 2048
 
-# Controls the maximum queue length of Unix-domain datagram sockets. High values may overload memory and are too redundant, too low values will cause packet loss. For servers, the optimal value is 256
-net.unix.max_dgram_qlen = 50
 
 # Set size of file handles and inode cache
 fs.file-max = 2097152
--- a/usr/lib/sysctl.d/99-cachyos-settings.conf
+++ b/usr/lib/sysctl.d/99-cachyos-settings.conf
@@ -74,9 +74,9 @@ net.ipv4.neigh.default.unres_qlen = 6
 net.ipv4.neigh.default.proxy_qlen = 96
 
 # Controls the minimum number of ARP entries at which to trigger purging of obsolete unused entries in the ARP table. Values that are too high increase memory and slow down ARP request processing. Too low - creates an unnecessary load on the system.
-net.ipv4.neigh.default.gc_thresh1 = 32
-net.ipv4.neigh.default.gc_thresh2 = 1024
-net.ipv4.neigh.default.gc_thresh3 = 2048
+net.ipv4.neigh.default.gc_thresh1 = 1024
+net.ipv4.neigh.default.gc_thresh2 = 2048
+net.ipv4.neigh.default.gc_thresh3 = 4096
 
 # Controls the ability to reuse sockets in the TIME-WAIT state. When set to 1, we reduce the latency of new connections
 net.ipv4.tcp_tw_reuse = 1
--- a/usr/lib/sysctl.d/99-cachyos-settings.conf
+++ b/usr/lib/sysctl.d/99-cachyos-settings.conf
@@ -85,7 +85,7 @@ net.ipv4.tcp_tw_reuse = 1
 net.ipv4.tcp_max_tw_buckets = 1440000
 
 # Controls the maximum number of pending connections for sockets. The higher the value, the more stable the network will be under high loads.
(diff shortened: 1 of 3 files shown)


Terromur:
Forgot to mention if necessary that nftables may not support `net.ipv4.tcp_fastopen = 3`, as it does not interact with it correctly and may not work properly. 

Perhaps this issue may be irrelevant, but it is necessary to have someone test this. [Link](https://serverfault.com/questions/1128629/nftables-possible-to-block-syn-packets-with-payload)

ventureoo:
Most of these settings are only needed for server workloads and switches. We've gotten similar requests before, and there's no point in using this on the desktop. So NACK from me.

Terromur:
> Most of these settings are only needed for server workloads and switches. We've gotten similar requests before, and there's no point in using this on the desktop. So NACK from me.

I'm not sure why only workloads, since these settings are configured for desktop computers as well, since the network works the same way as with workloads. I configured the settings optimally to how the settings should look and work for an average computer. 

Let's talk about that.

I can also remove settings that are not important, but only important settings that I think are important to add and I will explain why.

Here are, for example, the parameters that contributed. They also help to improve performance, but the expediency of adding such parameters is also useful for workloads for the most part. 
net.core.netdev_max_backlog
net.ipv4.tcp_timestamps

About the parameter
`net.ipv4.tcp_timestamps`.
I don't see the point of disabling it, as disabling it reduces the security and accuracy of TCP packets. Moreover, if the parameter is enabled, there will be no problems with stack overflow. So it is recommended to enable it. It is recommended to remove this parameter, as the default value is 1. [Link](https://docs.redhat.com/en/documentation/red_hat_enterprise_linux/8/html/monitoring_and_managing_system_status_and_performance/tuning-the-network-performance_monitoring-and-managing-system-status-and-performance#benefits-of-tcp-timestamps_tuning-the-network-performance)

Terromur:
Parameters aren't necessary.

PR #120, opened 2024-12-20
sysctl: Network Tunables Cleanup

1Naim:
There has been reports of slow connection speeds when using CachyOS that are not present when using other distributions. This has been bisected to our sysctl tunables. The other 2 commits serve as a cleanup to:
1. A potential security hazard
2. Keeping saner defaults

### DIFF
--- a/usr/lib/sysctl.d/99-cachyos-settings.conf
+++ b/usr/lib/sysctl.d/99-cachyos-settings.conf
@@ -48,7 +48,7 @@ net.ipv4.tcp_ecn = 1
 
 # Increase netdev receive queue
 # May help prevent losing packets
-net.core.netdev_max_backlog = 16384
+net.core.netdev_max_backlog = 4096
 
 # Disable TCP slow start after idle
 # Helps kill persistent single connection performance
--- a/usr/lib/sysctl.d/99-cachyos-settings.conf
+++ b/usr/lib/sysctl.d/99-cachyos-settings.conf
@@ -52,10 +52,6 @@ net.ipv4.tcp_fastopen = 3
 # TCP Enable ECN Negotiation for both outgoing and incoming connections
 net.ipv4.tcp_ecn = 1
 
-# TCP Reduce performance spikes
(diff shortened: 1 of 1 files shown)


ptr1337 (approved):

ventureoo:
I'd like to add that TFO is currently hardly used by any commonly used applications, so Firefox[1] and Chromium[2], and along with Electron/CEF-based applications, have long since dropped TFO support due to TLS issues and other potential security issues. [3][4]

[1] - https://bugzilla.mozilla.org/show_bug.cgi?id=1689604
[2] - https://issues.chromium.org/issues/40606022
[3] - https://web.archive.org/web/20230201161410/https://squeeze.isobar.com/2019/04/11/the-sad-story-of-tcp-fast-open/
[4] - https://blog.apnic.net/2021/07/05/tcp-fast-open-not-so-fast/

ventureoo:
Honestly there is a lot of conflicting information about `net.ipv4.tcp_timestamps`. On the one hand according to Red Hat it increases security, on the other hand there are many proponents of disabling net.ipv4.`tcp_timestamps` not only for performance reasons but also to reduce fingerprinting via TCP timestamps that expose machine time, even though now this time is randomized some seem to think this is insufficient: https://forums.whonix.org/t/do-ntp-and-tcp-timestamps-really-leak-your-local-time/7824/10

1Naim:
> I'd like to add that TFO is currently hardly used by any commonly used applications, so Firefox[1] and Chromium[2], and along with Electron/CEF-based applications, have long since dropped TFO support due to TLS issues and other potential security issues. [3][4]

All the more reason to just remove it from our settings :)

> Honestly there is a lot of conflicting information about `net.ipv4.tcp_timestamps`. On the one hand according to Red Hat it increases security, on the other hand there are many proponents of disabling net.ipv4.`tcp_timestamps` not only for performance reasons but also to reduce fingerprinting via TCP timestamps that expose machine time, even though now this time is randomized some seem to think this is insufficient: https://forums.whonix.org/t/do-ntp-and-tcp-timestamps-really-leak-your-local-time/7824/10

I used RH as the basis of knowledge for my commit. Whonix doesn't seem to be the best place to get information by here because the threat model is obviously different. We're better off sticking to the defaults for this until rigorously proven otherwise.

PR #124, opened 2025-01-03
udev: Use ntfs3 as the default driver for ntfs partitions

1Naim:
`ntfs` is classified as an unknown filesystem according to Linux and users can't directly mount such partitions directly. Instead, users must explicitly specify the filesystem type, e.g. `mount -t ntfs3 /dev/sdxY /mnt` to mount the partition successfully.

As a QoL change, adding this udev rule can use the `ntfs3` kernel driver by default without explicitly specifying it. Note that this can apparently confuse 3rd party tools and isn't recommended by the Arch wiki. (I'm not entirely sure what this "confusion" is. I guess we would need to have some NTFS users test this.)

Link: https://wiki.archlinux.org/title/NTFS#unknown_filesystem_type_'ntfs'

### DIFF
--- a/usr/lib/udev/rules.d/80-ntfs3-by-default.rules
+++ b/usr/lib/udev/rules.d/80-ntfs3-by-default.rules
@@ -1,0 +1,2 @@
+# Mount NTFS partitions as NTFS3 partitions without explicitly specifying NTFS3
+SUBSYSTEM=="block", ENV{ID_FS_TYPE}=="ntfs", ENV{ID_FS_TYPE}="ntfs3"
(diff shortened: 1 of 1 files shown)


1Naim:
> Wouldn't this break the current NTFS-3G installations? Because mounting without specifying a type when ntfs-3g is installed causes it to be used.

https://wiki.archlinux.org/title/NTFS-3G#Manual_mounting. Reading this, it seems like ntfs-3g partitions shouldn't break, and with what the udev rule is doing, the mount should be using `/usr/bin/mount.ntfs3` so it won't call `ntfs-3g`. This is all just speculation and nothing concrete. This is a valid concern so we should hold off on merging this for now. In the meantime, I guess we could add this to the wiki.

usr/lib/udev/rules.d/80-ntfs3-by-default.rules
```diff
@@ -0,0 +1,2 @@
+# Mount NTFS partitions as NTFS3 partitions without explicitly specifying NTFS3
+SUBSYSTEM=="block", ENV{ID_FS_TYPE}=="ntfs", ENV{ID_FS_TYPE}="ntfs3"
```
ventureoo, line 2: You can add this check for the time being until we figure out how safe is approach to use for all users, to make sure it only affects new installations without ntfs-3g.

```suggestion
SUBSYSTEM=="block", TEST!="/usr/bin/ntfs-3g", ENV{ID_FS_TYPE}=="ntfs", ENV{ID_FS_TYPE}="ntfs3"
```

ventureoo (approved):

1Naim:
One more thing to note is that it's commong for ntfs partitions to be marked as "dirty" making them unmountable unless fixed with `ntfsfix`. For this change to make sense, we would need to add https://aur.archlinux.org/packages/ntfsprogs-ntfs3 to our repos.

PR #129, opened 2025-02-05
fix: sbctl-batch-sign: Better handling of batch file signing by blacklisting known improper PE/COFF executables

1Naim:
Unfortunately, the prior version of the script caused massive issues for upstream[1][2] due to being poorly written and without any safeguards (mostly due to *a lot* of assumptions). The script is now hopefully better implemented due better protect against these issues by:
1. Actually filtering for unsigned files only this time; see  6d49e37ab94871cce421f9f508bce91f6027664f
2. Ignore ESP files from Microsoft and known files that are not proper PE executables; see 070ff71c21ec519aa97f76a7bc55506e3b9e9680 and eb4baca3911bc0fcac81e89e36ad6f8831e14d28
3. Protect against weird edgecases which may occur from the script due to empty output from either `sbctl` or `find`, and hypothetical filenames with spaces.

[1] https://github.com/Foxboron/sbctl/issues/414
[2] https://github.com/Foxboron/sbctl/issues/376

### DIFF
--- a/usr/bin/sbctl-batch-sign
+++ b/usr/bin/sbctl-batch-sign
@@ -12,5 +12,5 @@ fi
 if [ "$#" -eq 0 ]; then
     while IFS= read -r entries; do
         sbctl sign -s $entries
-    done < <(sort -u <(sbctl verify | grep 'signed' | cut -d' ' -f2) <(find /boot -maxdepth 1 -type f | grep vmlinuz))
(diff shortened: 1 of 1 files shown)


ptr1337 (approved):
Thanks, it works fine.

ptr1337 (approved):

PR #130, opened 2025-02-06
nvidia.conf: Enable switch for experimental better frame pacing

ptr1337:
Enable this globally, after it got tested by the community.

This should improve the frame pacing on VR Headset and non VRR high refresh rate monitors.

Example: 
At 240Hz each frame is expected every 4ms. But if a 1ms task—say, in the kernel or on the GSP—runs when a frame is about to be displayed, it can delay the rendering. Instead of a neat sequence at T+4ms, T+8ms, T+12ms, the frames might appear at T+4ms, T+9ms, T+12ms, etc. This shows how even small delays can shift frame timing, potentially impacting smooth display output.


ryzendew (approved):
i have tested this on my own and i can agree it's worth having on.

ventureoo (approved):
LGTM. It might be worth adding that this does not apply for PRIME configurations where display is controlled by the iGPU.

PR #131, opened 2025-02-08
sbctl-batch-sign: Export ESP_PATH and replace find workaround

1Naim:
Exporting this variable effectively gives us the same functionality as was given with the find workaround,
with the extra works that the listed files are actually ESP files, as validated by sbctl.

This removes the variable olympics we were doing as to not pass bogus lists to sbctl when signing.

This workaround is mainly for GRUB installs and old rEFInd installs. systemd-boot installs and new-ish rEFInd
installs (post partition table changes) won't need this workaround, and functionality remains the same there.

Unfortunately, this will break weird installs that have weird paths set as their ESP, e.g. /efi. However,
we were already hardcoding /boot in the find workaround anyway, so I'd like to believe that this won't change
anything. CachyOS only supports installs with /boot as their ESP at the moment, and if it ever occurs where
we need to support other ESP paths, we can add a detection for that (or read user input) very easily.

Tested on my CachyOS VM with GRUB boot manager
image

### DIFF
--- a/usr/bin/sbctl-batch-sign
+++ b/usr/bin/sbctl-batch-sign
@@ -17,7 +17,8 @@ if [ "$#" -eq 0 ]; then
         # own keys alongside Microsoft's.
         # With that in mind, there's no need to sign MS ESP
         # files with our own keys.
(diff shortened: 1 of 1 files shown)


vnepogodin (changes requested):
add `set -e`

vnepogodin (approved):

PR #133, opened 2025-02-17
udev: Drop ntsync rule, handled now directly by the kernel

ptr1337:
Upstream will add a patch, which makes the module automatically the correct permissions.

This has been applied to the RC (6.14.rc3) and 6.13.3 kernel. As soon they are available we can merge this.


ventureoo (approved):
LGTM. What's about LTS kernel?

1Naim:
> LGTM. What's about LTS kernel?

The related commit was cherry-picked in-tree what the patchset hasn't been synced yet.

1Naim (approved):

PR #135, opened 2025-02-22
dlss: add script to switch to latest dlss preset

ptr1337:
This adds script, which can be used in Steam as launch options.

The `dlss-swapper` does set the dlss preset to the latest available, and also update the dlls via ngx-updater.
NGX Updater is currently not working reliable at all games, and can result into crashes.

The `dlss-swapper-dll` script, only sets the preset to latest. This can be used by users, when they manually switch the dll's in the games directory.
Till ngx updater is working reliable, we likely want to keep the second script too, to make it easier for the user.

Im not sure, how we should call the scripts, so im open for suggestions.

Thanks to @ferrreo for bringing this script up :)

### DIFF
--- a/usr/bin/dlss-swapper
+++ b/usr/bin/dlss-swapper
@@ -1,0 +1,13 @@
+#!/usr/bin/bash
+
+# This forces Nvidia DLSS to use the latest preset for SR, RR and framegen + updates the dlss dlls via ngx
+
+export PROTON_ENABLE_NGX_UPDATER=1
+export DXVK_NVAPI_DRS_NGX_DLSS_RR_OVERRIDE=on
+export DXVK_NVAPI_DRS_NGX_DLSS_SR_OVERRIDE=on
+export DXVK_NVAPI_DRS_NGX_DLSS_FG_OVERRIDE=on
+export DXVK_NVAPI_DRS_NGX_DLSS_RR_OVERRIDE_RENDER_PRESET_SELECTION=render_preset_latest
+export DXVK_NVAPI_DRS_NGX_DLSS_SR_OVERRIDE_RENDER_PRESET_SELECTION=render_preset_latest
+
+# applied variables, now execute the rest of the command
+"$@"
--- a/usr/bin/dlss-swapper
+++ b/usr/bin/dlss-swapper
@@ -10,4 +10,4 @@ export DXVK_NVAPI_DRS_NGX_DLSS_RR_OVERRIDE_RENDER_PRESET_SELECTION=render_preset
 export DXVK_NVAPI_DRS_NGX_DLSS_SR_OVERRIDE_RENDER_PRESET_SELECTION=render_preset_latest
 
 # applied variables, now execute the rest of the command
-"$@"
+exec "$@"
--- a/usr/bin/dlss-swapper-dll
+++ b/usr/bin/dlss-swapper-dll
@@ -1,0 +1,12 @@
+#!/usr/bin/bash
+
+# This forces Nvidia DLSS to use the latest preset for SR, RR and framegen + skips ngx updater
+
+export DXVK_NVAPI_DRS_NGX_DLSS_RR_OVERRIDE=on
+export DXVK_NVAPI_DRS_NGX_DLSS_SR_OVERRIDE=on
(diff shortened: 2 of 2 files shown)


notpentadactyl:
The names are probably not the best, but I have no better idea either. Maybe someone else has a good idea.

usr/bin/dlss-swapper
```diff
@@ -0,0 +1,13 @@
...
+
+export PROTON_ENABLE_NGX_UPDATER=1
+export DXVK_NVAPI_DRS_NGX_DLSS_RR_OVERRIDE=on
+export DXVK_NVAPI_DRS_NGX_DLSS_SR_OVERRIDE=on
+export DXVK_NVAPI_DRS_NGX_DLSS_FG_OVERRIDE=on
+export DXVK_NVAPI_DRS_NGX_DLSS_RR_OVERRIDE_RENDER_PRESET_SELECTION=render_preset_latest
+export DXVK_NVAPI_DRS_NGX_DLSS_SR_OVERRIDE_RENDER_PRESET_SELECTION=render_preset_latest
+
+# applied variables, now execute the rest of the command
+"$@"
```
ventureoo, line 13:
```suggestion
exec "$@"
```

usr/bin/dlss-swapper-dll
```diff
@@ -0,0 +1,12 @@
...
+# This forces Nvidia DLSS to use the latest preset for SR, RR and framegen + skips ngx updater
+
+export DXVK_NVAPI_DRS_NGX_DLSS_RR_OVERRIDE=on
+export DXVK_NVAPI_DRS_NGX_DLSS_SR_OVERRIDE=on
+export DXVK_NVAPI_DRS_NGX_DLSS_FG_OVERRIDE=on
+export DXVK_NVAPI_DRS_NGX_DLSS_RR_OVERRIDE_RENDER_PRESET_SELECTION=render_preset_latest
+export DXVK_NVAPI_DRS_NGX_DLSS_SR_OVERRIDE_RENDER_PRESET_SELECTION=render_preset_latest
+
+# applied variables, now execute the rest of the command
+"$@"
```
ventureoo, line 12:
```suggestion
exec "$@"
```

ventureoo:
Do we really need two scripts for that? I think to skip NGX update we can just add some option in the first script.

vnepogodin:
Yeah the only diff is ngx update env

ptr1337:
> Do we really need two scripts for that? I think to skip NGX update we can just add some option in the first script.

How?

1Naim:
> > Do we really need two scripts for that? I think to skip NGX update we can just add some option in the first script.
> 
> How?

Not the best impl. but you can add a condition to check for `--dll` and if that's available, don't export that updater env

```sh
if [ "$1" != "--dll" ]; then
    export ENV=VAR
fi
```

ventureoo:
> > > Do we really need two scripts for that? I think to skip NGX update we can just add some option in the first script.
> > 
> > 
> > How?
> 
> Not the best impl. but you can add a condition to check for `--dll` and if that's available, don't export that updater env
> 
> ```shell
> if [ "$1" != "--dll" ]; then
>     export ENV=VAR
> fi
> ```

This will not work together with $@ in exec. You need to shift positional arguments in that case. Right implementation is like that:
```diff
diff --git a/usr/bin/dlss-swapper b/usr/bin/dlss-swapper
index 84339aa..82a94a4 100755
--- a/usr/bin/dlss-swapper
+++ b/usr/bin/dlss-swapper
@@ -9,5 +9,12 @@ export DXVK_NVAPI_DRS_NGX_DLSS_FG_OVERRIDE=on
 export DXVK_NVAPI_DRS_NGX_DLSS_RR_OVERRIDE_RENDER_PRESET_SELECTION=render_preset_latest
 export DXVK_NVAPI_DRS_NGX_DLSS_SR_OVERRIDE_RENDER_PRESET_SELECTION=render_preset_latest

+case "$1" in
+    --no-update)
+        export PROTON_ENABLE_NGX_UPDATER=0
+        shift;;
+    *);;
+esac
+
 # applied variables, now execute the rest of the command
 exec "$@"
```

ptr1337:
Merging for now, if someone has a better idea, please PR

PR #136, opened 2025-02-26
Variables to fix issues, help with performance and reduce latency

dougg0k:
It a script that can fix issues, help with performance and reduce latency.

Usually it's all scattered, this could help.

### DIFF
--- a/usr/bin/game-variables.sh
+++ b/usr/bin/game-variables.sh
@@ -32,7 +32,7 @@ export LD_PRELOAD=""
 
 # Reduce input lag, in exchange of a bit lower fps
 # https://github.com/HansKristian-Work/vkd3d-proton/issues/2273
-export VKD3D_SWAPCHAIN_LATENCY_FRAMES=2
+# export VKD3D_SWAPCHAIN_LATENCY_FRAMES=2
 
 # May work like DXVK numBackBuffers, which low value could reduce latency in some way.
 # export VKD3D_SWAPCHAIN_IMAGES=1
--- a/usr/bin/game-variables.sh
+++ b/usr/bin/game-variables.sh
@@ -1,0 +1,138 @@
+#!/usr/bin/bash
+
+# Uncomment or Comment what you might want to use or not.
+
+############################################################################
+
+# Enable NTSync (if available)
+# Can result in smoother experience and in some cases increased performance
+# export PROTON_USE_NTSYNC=1
+
+# Enable FSync (recommend to disable if using any other)
+export PROTON_NO_FSYNC=0
+export WINEFSYNC_FUTEX2=0
+export WINEFSYNC=0
+
+# Disable ESync
+export PROTON_NO_ESYNC=1
+export WINEESYNC=0
+
+# Disable Wine Sync and Fast Sync
+export WINE_DISABLE_FAST_SYNC=1
+export WINEESYNC=0
+export PROTON_USE_WINESYNC=0
+export PROTON_USE_WINED3D=0
+
+# Enable Native Wayland Driver (if available)
+# export PROTON_ENABLE_WAYLAND=1
+
+# Fix stuttering due to Steam overlay
+export LD_PRELOAD=""
+
+# Reduce input lag, in exchange of a bit lower fps
+# https://github.com/HansKristian-Work/vkd3d-proton/issues/2273
+export VKD3D_SWAPCHAIN_LATENCY_FRAMES=2
+
+# May work like DXVK numBackBuffers, which low value could reduce latency in some way.
+# export VKD3D_SWAPCHAIN_IMAGES=1
+
+# Unsafe speed hack on NVIDIA. May or may not give a significant performance uplift.
+# no_upload_hvv - May free up vital VRAM in certain critical situations, at cost of lower GPU performance.
+# nodxr - May help disable raytracing if the implementaion are not working well with linux and forcing.
+# You can use additional config by separating in the same line with a comma.
+export VKD3D_CONFIG=force_static_cbv
+
+# Set realtime priority for the wineserver which can result in improved performance
+export STAGING_RT_PRIORITY_SERVER=90
+export STAGING_RT_PRIORITY_BASE=90
+
+# Enable Enable Hardware Accelerated GPU Scheduling
+# Can result in improved performance
+export WINEHAGS=1
+export WINE_DISABLE_HARDWARE_SCHEDULING=0
+
+# Prefer SDL instead of Hidraw
+# May help increase compatibility with different devices
+export PROTON_PREFER_SDL=1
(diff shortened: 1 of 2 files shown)


loathingKernel:
The script seems to assume that it is meant to be edited, and that is suggested by the aggressiveness of some options that are not good defaults, even if they might help with performance. I don't think something like this should live in a read-only location such as `/usr/bin` that requires root access to edit. I think you should either find a way to read these from a user-local location where the script will create a commented skeleton file, or it should live under the user's directory.

dougg0k:
How do you suggest to add the file under user directory?

Create a folder `~/` inside your repo?

Besides `VKD3D_SWAPCHAIN_LATENCY_FRAMES` I dont see anything else being bad values to be defaults. Which ones do you mean?

loathingKernel:
> How do you suggest to add the file under user directory?

One way would be to have the script dump another bash script in `~/.local/shaare/some_location` if it doesn't exist with everything commented out, or at least the bare minimum of defaults, the entry point script can check for the dumped script exists at that location and source it in subsequent runs,

> Also, this were meant to be an optional script. And this script could help prevent many questions being asked.

The option to enable it still requires root access to be edited

>  Which ones do you mean?

I will remove them individually sortly

dougg0k:
There shouldnt be any need to remove anything, i've used them here, works very well.

If someone are going to use the script, they should have the options there to use as they want. Regardless of your good defaults.

usr/bin/game-variables.sh
```diff
@@ -0,0 +1,139 @@
...
+
+# Disable ESync
+export PROTON_NO_ESYNC=1
+export WINEESYNC=0
+
+# Disable Wine Sync and Fast Sync
+export WINE_DISABLE_FAST_SYNC=1
+export WINEESYNC=0
+export PROTON_USE_WINESYNC=0
+export PROTON_USE_WINED3D=0
```
loathingKernel, line 25: This section enforces options that the user might want to override per-case, this is meant to be part of `game-performance` which typically comes after the user sets some env variables, at the very least the script should respect these env variables if they are already set instead of ignoring them and enforcing its own.

usr/bin/game-variables.sh
```diff
@@ -0,0 +1,139 @@
...
+# export VKD3D_SWAPCHAIN_LATENCY_FRAMES=2
+
+# May work like DXVK numBackBuffers, which low value could reduce latency in some way.
+# export VKD3D_SWAPCHAIN_IMAGES=1
+
+# Unsafe speed hack on NVIDIA. May or may not give a significant performance uplift.
+# no_upload_hvv - May free up vital VRAM in certain critical situations, at cost of lower GPU performance.
+# nodxr - May help disable raytracing if the implementaion are not working well with linux and forcing.
+# You can use additional config by separating in the same line with a comma.
+export VKD3D_CONFIG=force_static_cbv
```
loathingKernel, line 44: As your comment says this described as ["unsafe speed hack for Nvidia"](https://github.com/HansKristian-Work/vkd3d-proton?tab=readme-ov-file#environment-variables) and just sits here without any checks if it should be applied (for example when using AMD).

usr/bin/game-variables.sh
```diff
@@ -0,0 +1,139 @@
...
+export STAGING_RT_PRIORITY_BASE=90
+
+# Enable Enable Hardware Accelerated GPU Scheduling
+# Can result in improved performance
+export WINEHAGS=1
+export WINE_DISABLE_HARDWARE_SCHEDULING=0
+
+# Prefer SDL instead of Hidraw
+# May help increase compatibility with different devices
+export PROTON_PREFER_SDL=1
```
loathingKernel, line 57: This can cause issues with gaming wheels and other peripherals. It should not be the default.

usr/bin/game-variables.sh
```diff
@@ -0,0 +1,139 @@
...
+export STAGING_SHARED_MEMORY=1
+export STAGING_WRITECOPY=1
+
+# Load shared objects immediately for better first time latency
+# https://wiki.archlinux.org/title/Gaming#Load_shared_objects_immediately_for_better_first_time_latency
+export LD_BIND_NOW=1
+
+# Use latest shader level and model
+export VKD3D_FEATURE_LEVEL=12_2
+export VKD3D_SHADER_MODEL=6_6
```
loathingKernel, line 88: Not true for older GPUs, sometimes these need to be adjusted, they should not be defaults.

usr/bin/game-variables.sh
```diff
@@ -0,0 +1,139 @@
...
+# May help if you are using dxvk async version
+# This version can help in some edge cases, but most of the time, it's recommended to stay in the default.
+# export DXVK_ASYNC=1
+# export DXVK_GPLASYNCCACHE=1
+
+# May help 32-bit games with address space limitations
+export WINE_LARGE_ADDRESS_AWARE=1
+
+# Disable valve fullscreen hack and rely on monitor scaling instead
+export WINE_DISABLE_FULLSCREEN_HACK=1
```
loathingKernel, line 112: Why??? Just why? This will cause the desktop resolution to change, and mess things up.

usr/bin/game-variables.sh
```diff
@@ -0,0 +1,139 @@
...
+# May help fix potential audio issues with the new implementation
+# export PROTON_AUDIO_CONVERT=0
+# export PROTON_AUDIO_CONVERT_BIN=0
+# export PROTON_VIDEO_CONVERT=0
+# export PROTON_DEMUX=0
+
+# Disable logs
+# May help reduce resource usage
+export PROTON_LOG=0
+export WINEDEBUG=-all
```
loathingKernel, line 137: Very bad default for identifying issues. It should at least respect what is already set.

usr/bin/game-variables.sh
```diff
@@ -0,0 +1,139 @@
...
+export WINE_DISABLE_FAST_SYNC=1
+export WINEESYNC=0
+export PROTON_USE_WINESYNC=0
+export PROTON_USE_WINED3D=0
+
+# Enable Native Wayland Driver (if available)
+# export PROTON_ENABLE_WAYLAND=1
+
+# Fix stuttering due to Steam overlay
+export LD_PRELOAD=""
```
loathingKernel, line 31: This also disables the steam overlay IIRC, once the issue is fixed this is going to make people wonder why their steam overlay is not working.

loathingKernel (changes requested):
This is a good documentation and easy access to commonly used and useful variables. But this is not the way to do it at all. The whole `game-variables` script should become a commented-by-default here-script that is written in the user's directory where they can edit it and subsequently sourced. It should not require any kind of root access to be edited and it also should respect the env variables if they are already set.

dougg0k:
I can answer each, but wont. You are free to modify as you want or close the pr.

PR #137, opened 2025-03-05
nvidia.conf: Enable GSP Firmware again

ptr1337:
NVIDIA has recently fixed the GSP Firmware stutters and also performance in some scenarios.

Since this is now not anymore a major issue, we should enable the GSP firmware again and following nvidia's defaults.
Users, which might still run into issues can change it manually back.


ventureoo (changes requested):
There's no need to leave a comment here. We already have info about it on the wiki, and user is supposed to create an override file in ``/etc``, not edit existing one in ``/usr/lib``.

ventureoo (approved):

PR #138, opened 2025-03-10
udev: Use arch linux wiki zram rules

EliasTheGrandMasterOfMistakes:
from: https://wiki.archlinux.org/title/Zram#Optimizing%20swap%20on%20zram


usr/lib/udev/rules.d/30-zram.rules
```diff
@@ -9,6 +9,12 @@ TEST!="/dev/zram0", GOTO="zram_end"
 # the page cache as hermetic as possible, because otherwise it is "expensive"
 # to read data from disk again. At the same time, uncompressing pages from ZRAM
 # is not as expensive and is usually very fast on modern CPUs.
-SYSCTL{vm.swappiness}="150"
+SYSCTL{vm.swappiness}="180"
+
+# Based on https://wiki.archlinux.org/title/Zram#Optimizing%20swap%20on%20zram 
+SYSCTL{vm.watermark_boost_factor}="0"
+SYSCTL{vm.watermark_scale_factor}="125"
+SYSCTL{vm.page-cluster}="0"
```
1Naim, line 17: This is already set via sysctl

usr/lib/udev/rules.d/30-zram.rules
```diff
@@ -9,6 +9,12 @@ TEST!="/dev/zram0", GOTO="zram_end"
 # the page cache as hermetic as possible, because otherwise it is "expensive"
 # to read data from disk again. At the same time, uncompressing pages from ZRAM
 # is not as expensive and is usually very fast on modern CPUs.
-SYSCTL{vm.swappiness}="150"
+SYSCTL{vm.swappiness}="180"
+
+# Based on https://wiki.archlinux.org/title/Zram#Optimizing%20swap%20on%20zram 
+SYSCTL{vm.watermark_boost_factor}="0"
```
1Naim, line 15: This is already disabled in the kernel.

1Naim (changes requested):

usr/lib/udev/rules.d/30-zram.rules
EliasTheGrandMasterOfMistakes: is just vm.page-cluster? because vm.watermark_scale_factor i can't found in usr/lib/sysctl.d

1Naim: Yeah, that isn't set yet. I think I'm generally fine with it, considering some other distro is shipping it. Can you move it to the sysctl file tho? It's more appropriate to place it there.

@ventureoo thoughts on settings `vm.watermark_scale_factor`?

ptr1337: We have removed this since this increased the ram usage (cache) like x2

EliasTheGrandMasterOfMistakes: > We have removed this since this increased the ram usage (cache) like x2

sorry, this means that is best to not set it? and just increase zram in 180?

EliasTheGrandMasterOfMistakes: or this parameter can be set when just using zram?

EliasTheGrandMasterOfMistakes:
> The zram rules are handled here: https://github.com/CachyOS/CachyOS-Settings/blob/master/usr/lib/udev/rules.d/30-zram.rules
> 
> The reason for this, is to only apply the changes, when zram is found.

thank you, i has added in this file 30-zram.rules

usr/lib/udev/rules.d/30-zram.rules
ptr1337: We currently set swapiness to 150, as you see in the below comment. Im unsure, if 180 will make a huge difference?

You can also test yourself with the above values - they go way over the roof. We have debugged this "issue" very long time, till we found this is coming from there.

This was removed in commit [`451836f` (#55)](https://github.com/CachyOS/CachyOS-Settings/pull/55/commits/451836f2955f2ca54ee34b3c2d03166acb9a63a6)

EliasTheGrandMasterOfMistakes: in my test looks that zram usage is more than using 150, idk if this are a placebo effect, but this makes zram be used more early when running applications that uses high ram. i has reduced zram priority from 100 to 0 too.
I think that can be good if has more results, but for now can be just priority or a placebo

usr/lib/udev/rules.d/30-zram.rules
EliasTheGrandMasterOfMistakes: thank you, i has made a forced push to remove

usr/lib/udev/rules.d/30-zram.rules
ventureoo: > Yeah, that isn't set yet. I think I'm generally fine with it, considering some other distro is shipping it. Can you move it to the sysctl file tho? It's more appropriate to place it there.
> 
> @ventureoo thoughts on settings `vm.watermark_scale_factor`?

To be honest, I'm not sure. Basiclly that is just makes kswapd less aggressive when RAM usage is low, and in turn it actually increases memory consumption since most of unused pages stay in memory. This is positive effect for CPU in case of ZRAM and makes some users happy, because they don't like earlier swap usage for whatever reason, but on other hand creates discomfort for other as the increase can be quite large [1]. Probably we just need to pick a some reasonably balanced value, as we have done with other non-obvious parameters.

I also want to separately note that kswapd is not only about swap, but also about freeing unused file pages, i.e. flushing page cache, which is in my view the main reason for the increased consumption here even when there are no pages that actually need to be swapped into ZRAM.

[1] - https://unix.stackexchange.com/questions/729275/why-does-setting-the-vm-watermark-scale-factor-property-increase-ram-consumption

usr/lib/udev/rules.d/30-zram.rules
ventureoo: > This is already disabled in the kernel.

Well, actually it's not. It's just we have patch that enables it by default, but user can use other kernels.

1Naim: > > This is already disabled in the kernel.
> 
> Well, actually it's not. It's just we have patch that enables it by default, but user can use other kernels.

Yeah, that's what I meant. Do we really want to care about other kernels though? It's far more likely that outside users would use the kernel more than our settings. There's a high chance that their already using our kernel too when they opt to use our settings. Other kernels such as zen and tkg also disable this via patch too AFAIK.

PR #146, opened 2025-06-02
sbctl-batch-sign: Remove unnecessary root check

1Naim:
We're already checking if the user isn't root before the main body of the script, so there is no need to wrap it again with a check if the user is root

### DIFF
--- a/usr/bin/sbctl-batch-sign
+++ b/usr/bin/sbctl-batch-sign
@@ -15,18 +15,15 @@ if [ "$(id -u)" -ne 0 ]; then
     exit 1
 fi
 
-if [ "$#" -eq 0 ]; then
-    export ESP_PATH=/boot
-
-    sbctl verify 2>/dev/null | awk '/✗/ {print $2}' | while IFS= read -r entry; do
-        # We expect users who use this script to enroll their
-        # own keys alongside Microsoft's.
-        # With that in mind, there's no need to sign MS ESP
-        # files with our own keys.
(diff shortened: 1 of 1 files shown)


ptr1337 (approved):

usr/bin/sbctl-batch-sign
```diff
@@ -15,18 +15,15 @@ if [ "$(id -u)" -ne 0 ]; then
...
-        # With that in mind, there's no need to sign MS ESP
-        # files with our own keys.
-        if [[ "$entry" =~ ^.*/EFI/(Microsoft|Windows) || "$entry" == *.mui || "$entry" == *.dll
-            || "$entry" =~ ^/boot/grub ]]; then
-            continue
-        fi
-        sbctl sign -s "$entry"
-    done
-fi
+export ESP_PATH=/boot
```
ventureoo, line 18: That's probably better than hardcoding.
```suggestion
BOOT_PATH="$(bootctl -p 2>/dev/null)"
export ESP_PATH="${BOOT_PATH:-/boot}"
```

1Naim: Well no, the reason I'm hardcoding this in the first place is because on GRUB, ESP_PATH is /boot/efi, so sbctl won't be able to find the kernel images when doing `sbctl verify` and by extension, when running this script. If we drop support for GRUB then we can remove this export entirely.

ventureoo (approved):

PR #147, opened 2025-06-03
udev: Add rule to prevent audio cracking on some hardware

ventureoo:
I received a number of reports, and decided to make a udev rule to prevent such issues with condition that we don't degrade battery life of the devices. 

Needs testing on desktop PCs.

PR #149, opened 2025-06-07
udev: Do not disable powersaving for NVIDIA audio devices

ventureoo:
Somehow, disabling power saving for NVIDIA audio device also prevents the dGPU from going into sleep mode.

See: https://github.com/CachyOS/CachyOS-Settings/pull/147#issuecomment-2952595836

### DIFF
--- a/usr/lib/udev/rules.d/20-audio-pm.rules
+++ b/usr/lib/udev/rules.d/20-audio-pm.rules
@@ -5,4 +5,5 @@ SUBSYSTEM=="power_supply", ENV{POWER_SUPPLY_ONLINE}=="0", TEST=="/sys/module/snd
     RUN+="/bin/sh -c 'echo Y > /sys/module/snd_hda_intel/parameters/power_save_controller'"
 
 SUBSYSTEM=="power_supply", ENV{POWER_SUPPLY_ONLINE}=="1", TEST=="/sys/module/snd_hda_intel", \
-    RUN+="/bin/sh -c 'echo N > /sys/module/snd_hda_intel/parameters/power_save_controller'"
+    RUN+="/bin/sh -c 'echo $(grep -w -q 0x10de /sys/module/snd_hda_intel/drivers/*/*/vendor \
+        && echo Y || echo N) > /sys/module/snd_hda_intel/parameters/power_save_controller'"


ventureoo:
> Happy to confirm that this fixes the issue :)

Nice, thanks for the quick report and testing!

> I'm equally surprised that RTD3 is THAT fragile...

Yeah, it's a bit upsetting, but as it is...

asumagic:
FWIW I independently found #147 broke D3 for my GPU too - except I have an **AMD** 7900XT, not an NVIDIA GPU. So it affected AMD GPUs as well...

Mentioning this for the changelog, but that's not super important :)

asumagic:
Oh wait -- I didn't actually try this PR, only removed the original file (as I don't use intel audio anyway). The workaround seems to exclude only NVIDIA GPUs, this is wrong for the reason above...

I don't have any such hardware to test, but maybe it might cause issues for intel dGPU HDMI audio?

ventureoo:
> Oh wait -- I didn't actually try this PR, only removed the original file (as I don't use intel audio anyway). The workaround seems to exclude only NVIDIA GPUs, this is wrong for the reason above...
> 
> I don't have any such hardware to test, but maybe it might cause issues for intel dGPU HDMI audio?

Can you please share output of ``cat /sys/module/snd_hda_intel/drivers/*/*/class``?

asumagic:
> > _snip_
> 
> I see, thank you. Could you also check if the problem is present if change values as in this diff:
> 
> ```diff
> diff --git a/usr/lib/modprobe.d/20-audio-pm.conf b/usr/lib/modprobe.d/20-audio-pm.conf
> index 761a2fa..1a0d1b9 100644
> --- a/usr/lib/modprobe.d/20-audio-pm.conf
> +++ b/usr/lib/modprobe.d/20-audio-pm.conf
> @@ -2,4 +2,4 @@
>  # running on battery power. This is needed because it prevents audio cracks on
>  # some hardware. If device starts working on battery power, same named udev rule
>  # re-enables power saving.
> -options snd_hda_intel power_save_controller=N
> +options snd_hda_intel power_save=0
> diff --git a/usr/lib/udev/rules.d/20-audio-pm.rules b/usr/lib/udev/rules.d/20-audio-pm.rules
> index fa4f11c..b0f88a0 100644
> --- a/usr/lib/udev/rules.d/20-audio-pm.rules
> +++ b/usr/lib/udev/rules.d/20-audio-pm.rules
> @@ -2,7 +2,7 @@
>  # running on battery power. This is needed because it prevents audio cracks on
>  # some hardware.
>  SUBSYSTEM=="power_supply", ENV{POWER_SUPPLY_ONLINE}=="0", TEST=="/sys/module/snd_hda_intel", \
> -    RUN+="/bin/sh -c 'echo Y > /sys/module/snd_hda_intel/parameters/power_save_controller'"
> +    RUN+="/bin/sh -c 'echo 0 > /sys/module/snd_hda_intel/parameters/power_save'"
> 
>  SUBSYSTEM=="power_supply", ENV{POWER_SUPPLY_ONLINE}=="1", TEST=="/sys/module/snd_hda_intel", \
> -    RUN+="/bin/sh -c 'echo $(grep -w -q 0x10de /sys/module/snd_hda_intel/drivers/*/*/vendor && echo Y || echo N) > /sys/module/snd_hda_intel/parameters/power_save_controller'"
> +    RUN+="/bin/sh -c 'echo 10 > /sys/module/snd_hda_intel/parameters/power_save'"
> ```

Applied this and rebooted, the problem seems to be solved.

(also double-checked my `/etc` for modprobe/udev to make sure I'm not overriding anything there -- seems good too.)

ventureoo:
> > > _snip_
> > 
> > 
> > I see, thank you. Could you also check if the problem is present if change values as in this diff:
> > ```diff
> > diff --git a/usr/lib/modprobe.d/20-audio-pm.conf b/usr/lib/modprobe.d/20-audio-pm.conf
> > index 761a2fa..1a0d1b9 100644
> > --- a/usr/lib/modprobe.d/20-audio-pm.conf
> > +++ b/usr/lib/modprobe.d/20-audio-pm.conf
> > @@ -2,4 +2,4 @@
> >  # running on battery power. This is needed because it prevents audio cracks on
> >  # some hardware. If device starts working on battery power, same named udev rule
> >  # re-enables power saving.
> > -options snd_hda_intel power_save_controller=N
> > +options snd_hda_intel power_save=0
> > diff --git a/usr/lib/udev/rules.d/20-audio-pm.rules b/usr/lib/udev/rules.d/20-audio-pm.rules
> > index fa4f11c..b0f88a0 100644
> > --- a/usr/lib/udev/rules.d/20-audio-pm.rules
> > +++ b/usr/lib/udev/rules.d/20-audio-pm.rules
> > @@ -2,7 +2,7 @@
> >  # running on battery power. This is needed because it prevents audio cracks on
> >  # some hardware.
> >  SUBSYSTEM=="power_supply", ENV{POWER_SUPPLY_ONLINE}=="0", TEST=="/sys/module/snd_hda_intel", \
> > -    RUN+="/bin/sh -c 'echo Y > /sys/module/snd_hda_intel/parameters/power_save_controller'"
> > +    RUN+="/bin/sh -c 'echo 0 > /sys/module/snd_hda_intel/parameters/power_save'"
> > 
> >  SUBSYSTEM=="power_supply", ENV{POWER_SUPPLY_ONLINE}=="1", TEST=="/sys/module/snd_hda_intel", \
> > -    RUN+="/bin/sh -c 'echo $(grep -w -q 0x10de /sys/module/snd_hda_intel/drivers/*/*/vendor && echo Y || echo N) > /sys/module/snd_hda_intel/parameters/power_save_controller'"
> > +    RUN+="/bin/sh -c 'echo 10 > /sys/module/snd_hda_intel/parameters/power_save'"
> > ```
> 
> Applied this and rebooted, the problem seems to be solved.
> 
> (also double-checked my `/etc` for modprobe/udev to make sure I'm not overriding anything there -- seems good too.)

Thanks, looks like this should be finally fixed.

sankix69:
I'm a bit confused - as far as I understand, this udev rule is supposed to set power_save to 0 when the laptop is charging, and to 10 when running on battery.

However, on my system, when the laptop is charging, the output of cat /sys/module/snd_hda_intel/parameters/power_save is 10, and when it's not charging, the output is 0.

Isn't this reversed?

PR #150, opened 2025-06-16
modprobe.d/nvidia: Remove modeset parameter

ventureoo:
All NVIDIA module packages in our repository are patched to enable modeset by default, so this is unnecessary, but can also cause issues on older driver versions such as 390.xx.

See: https://github.com/CachyOS/CachyOS-PKGBUILDS/commit/396ecc5a4a11768ff387baffe35824e473bc9b43


1Naim (approved):
Fine by me, the Arch package also already has this enabled by default.

ptr1337 (approved):

PR #157, opened 2025-07-24
refactor(game-performance): gaming-power-profile

Zeioth:
## Proposal
This little daemon pretends to solve the problem of game-performance not being accessible from flatpak, by automatically setting power mode to "performance" when a steam game is running. Once the game is not running anymore, performance will be back to "balanced".

This PR is to game-performance what ananicy-cpp is to gamemode, kinda.

## PROS
* It can be used with flatpak.
* It doesn't require to pass "game-performance" to every game where we want to enable it.

## CONS
* Another daemon running in background.
* Not super transparent with the user.

## More info
I haven't PR the service code itself. But it's pretty straightforward. 

```
touch ~/.config/systemd/user/gaming-power-profile.service
```
And paste
```
[Unit]
Description=Gaming Power Profile
After=graphical-session.target

[Service]
ExecStart=%h/.local/bin/gaming-power-profile
Restart=on-failure
Environment=PATH=%h/.local/bin:/usr/bin:/bin

[Install]
WantedBy=default.target
```

Then enable it with
```
systemctl --user daemon-reload
systemctl --user enable --now gaming-power-profile.service
```
If this PR is accepted, please, don't forget to add the service and delete game-performance.

### DIFF
--- a/usr/bin/gaming-power-profile
+++ b/usr/bin/gaming-power-profile
@@ -11,11 +11,10 @@ if ! command -v powerprofilesctl &>/dev/null; then
 fi
 
 #######################################
-# Returns: true  - If a game is running.
-#          false - Otherwise.
+# Game detection
 #######################################
 is_game_running() {
-    # Steam game is running — very specific
+    # Steam game is running
     pgrep -f "steamapps/common" >/dev/null && return 0
 
     # Lutris (native or Flatpak)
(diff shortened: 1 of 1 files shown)


Zeioth:
Maybe it would be easier to test if I publish it on AUR.

Zeioth:
This script has the obvious caveat of being too opinionated; 

If we set power to "balanced" when games are not running, that will prevent users from setting their own behavior. Like for example running performance mode permanently, which some users might preffer.

Zeioth:
This script have the obvious caveat of being too opinionated;

Because we are setting "balanced" while the user is not gaming, that will prevent any custom behavior users might prefer, like setting performance mode permanently, for example.

SoongVilda:
> updating the system

May I ask why would you enable it during system update?

Zeioth:
In the end I've decided to make a proper implementation on Rust.
https://github.com/Zeioth/power-rules-daemon

This is as much time as I have today. But my idea is to release it on cargo, so any distro can distribute it easily.

## New features
- Config file: Rules can now be user defined!
- Config file: Changes are applied on real time!
- User interaction detection: If a user manually changes the power profile on the desktop environment or terminal, the daemon is paused for n minutes (configurable!).

What do you think?

PR #158, opened 2025-08-02
profile.d: Set ELECTRON_OZONE_PLATFORM_HINT to Auto

ptr1337:
The issue on RTX Cards with obsidian appears to be fixed now. 

It should be now fine to export this for all users

### DIFF
--- a/etc/profile.d/electron-ozone.sh
+++ b/etc/profile.d/electron-ozone.sh
@@ -1,0 +1,1 @@
+export ELECTRON_OZONE_PLATFORM_HINT=auto


vnepogodin (approved):

azdanov (approved):

ventureoo:
> It should be now fine to export this for all users

This will not work properly for GNOME/KDE users, as applications run on them as separate user units that have their own set of environment variables. 

You need to create  a`/etc/environment.d` file that will add this variable for them.

PR #160, opened 2025-08-04
nvidia.conf: Enable S0ix Power Management

ptr1337:
This has been enabled by Suse and Fedora since a while. Lets also enable this for newer GPUs, which should yield longer battery time, when in sleep.

### DIFF
--- a/usr/lib/modprobe.d/nvidia.conf
+++ b/usr/lib/modprobe.d/nvidia.conf
@@ -30,7 +30,15 @@
 # shows how even small delays can shift frame timing, potentially impacting
 # smooth display output.
 #
+# NVreg_EnableS0ixPowerManagement=1 (default 0) Enables S0ix for the NVIDIA GPU:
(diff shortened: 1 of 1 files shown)


ventureoo:
> reducing battery drain—especially on laptops

Is there any source? Because the NVIDIA documentation is not so clear about this:

> S0ix-based suspend will consume more power than legacy S3 system suspend, but it will enter and exit suspend/resume more quickly.

(https://download.nvidia.com/XFree86/Linux-x86_64/580.65.06/README/powermanagement.html)

Generally, I don't mind,  just want to make sure there are no major regressions.

Fabback:
@ptr1337 
I've had no problem for a few month with suspend on my NVidia GTX 970, but since a few days it crashes again and displays a long journal full screen with no way out. After removing this parameter, then restarting, suspend seems to work again (i've only tried once after a 2 min inactivity)
Will this parameter revert back automatically on any update ?

PR #161, opened 2025-08-04
nvidia.conf: Remove RMIntrLockingMode && enable AggressiveVblank

ptr1337:
This is for the 580 series, when getting merged.

RMIntrLockingMode has been enabled OOB.

The new RmEnableAggressiveVblank=1 is a experimental flag:

> Implemented another feature that can reduce time spent in the interrupt top half for low latency display interrupts by deferring the work until later. This feature is experimental and disabled by default. This feature can be enabled by loading nvidia.ko with the `NVreg_RegistryDwords=RmEnableAggressiveVblank=1` kernel module parameter.

### DIFF
--- a/usr/lib/modprobe.d/nvidia.conf
+++ b/usr/lib/modprobe.d/nvidia.conf
@@ -16,19 +16,6 @@
 # management for Turing generation mobile cards, allowing the dGPU to be
 # powered down during idle time.
 #
-# NVreg_RegistryDwords=RMIntrLockingMode=1 (default 0) - enables experimental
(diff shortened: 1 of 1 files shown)


ventureoo:
It’s probably better to split this into two commits, as we have not yet tested AggressiveVblank properly.

ptr1337:
Has been merged now partially here:
https://github.com/CachyOS/CachyOS-Settings/commit/a559ecccf0844dc1f4a81bdedf3b5e15899304b6

AggresiveVblank is problematic according some users. Closing.

Issue #177, opened 2025-11-09
Dont use 99- prefix, to allow users to overwrite config without  breaking the 2-digit naming scheme


laundmo:
I had the need to overwrite some of the sysctl config set by this, but by choosing the 99- prefix you've made it so I can't add a higher priority file with my own preferences.

I now have this:
```
.rw-r--r--  2,5k root 31 Okt 13:03  99-cachyos-settings.conf
.rw-r--r--    55 root  9 Nov 20:39  999-custom.conf
.rw-r--r--  1,8k root  9 Nov 20:38  9999-coredump.conf
.rw-r--r--    25 root  9 Nov 20:32  99999-ebpf.conf
```
Thats pretty inconvenient...

Note: both `sysctl --system` and `systemd-sysctl.service` have higher numbers take precedence.

damachine:
Perhaps the CachyOS settings file should use a lower number, like `80-cachyos-settings.conf` or `90-cachyos-settings.conf`. This would allow users to override the settings with `99-custom.conf` while maintaining the default two-letter naming scheme. As a workaround, you could try doing it manually first if necessary.

PR #184, opened 2025-11-16
Decrease sysctl config prefix

ventureoo:
Closes: https://github.com/CachyOS/CachyOS-Settings/issues/177


1Naim (commented):
Let's do 70 instead to make sure all the default configs are read first.

ventureoo:
> Let's do 70 instead to make sure all the default configs are read first.

Yeah, it makes sense. Done.

PR #185, opened 2025-11-18
Update network performance 99-cachyos-settings.conf

SoongVilda:
I have been testing these for about a week. I had some inaccurate values, but it seems promising now.

Rustdesk can achieve usable gaming at 80-100 FPS on a LAN connection. Sunshine + Moonlight also works perfectly, even at a 150 Mbps bitrate with H.265, with no drops or lags.

Results over WAN (Tailscale) are pretty similar when direct P2P is achieved and both sides have a decent network.

I believe my personal tweaks would be useful for CachyOS.

### DIFF
--- a/usr/lib/sysctl.d/99-cachyos-settings.conf
+++ b/usr/lib/sysctl.d/99-cachyos-settings.conf
@@ -98,7 +98,8 @@ net.ipv4.tcp_mtu_probing = 1
 net.ipv4.tcp_slow_start_after_idle = 0
 
 # --- 9. Aggressive Keepalives (NAT Persistence) ---
-# Reduces the time before keepalive packets are sent (Default: 2 hours -> 60s).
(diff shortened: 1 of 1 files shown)


1Naim:
> Rustdesk can achieve usable gaming at 80-100 FPS on a LAN connection. Sunshine + Moonlight also works perfectly, even at a 150 Mbps bitrate with H.265, with no drops or lags

Do you have any benchmarks against stock settings?

VolRencs:
Sounds interesting, but to understand how it really is you need tests in different situations. I believe that there is no way without benchmarks. Also because of these changes there may be small problems for people who have less than 4GB of RAM, correct me if I'm wrong.

damachine:
Interesting changes. I know from experience that network benchmarks can be very time-consuming, as many external factors come into play. Nevertheless, out of interest, I will conduct a before-and-after comparison of network performance.

PR #192, opened 2025-12-06
feat: Configure Qdisc and enable BBR for TCP

damachine:
Added settings for Qdisc and TCP congestion control. 
Without both settings, BBR3 won't be used even with the kernel patch applied.

Reports here:
https://discuss.cachyos.org/t/bbr3-not-working/12908


damachine:
I closed it because: if someone is using a kernel without the BBr3 patch, they should receive an error message in the log. Even though `tcp_bbr` has been included in the kernel for a long time (BBR v1 was released in kernel 4.9 in 2016). I'll leave it open for review and feedback. I don't know if the legacy module (up to BBR v1) is sufficient as a fallback.

PR #194, opened 2025-12-29
Remove NVreg_DynamicPowerManagement option

francoism90:
See: https://download.nvidia.com/XFree86/Linux-x86_64/465.19.01/README/dynamicpowermanagement.html

```
Option "NVreg_DynamicPowerManagement=0x03"
This is the default setting.

For Ampere or later notebooks with supported configurations, this value translates to fine-grained power control. For pre-Ampere notebooks, this value disables runtime D3 power management features.
```

I believe it does uses `0x02` by default on pre-Ampere devices.


francoism90:
It's a bit confusing how NVIDIA wrote the docs, but thanks for the explanation. :)

ventureoo (approved):

PR #196, opened 2026-01-15
Add mousewheel scroll speed quirk for Logitech G502 in libinput

raphipod:
This is a temporary fix for a problem related to the "hid-logitech-hidpp" driver, which causes the mousewheel to send too many events in a short period causing the wheel to be moved too fast.

This affects CS2, Minecraft and Hytale in my testing. Both in Hytale and Minecraft, when using the mousewheel you should be able to scroll slot by slot, not skip 10 of them. I didn't test CS2 yet, but I had problems with not being able to switch between weapons correctly because it was scrolling too fast.

This is a temporary fix - it basically ignores the high-resolution inputs of the mousewheel and makes scrolling in game hotbars/inventories more pleasant again.

In my testing, it does not interfere with the scrolling speed in any other app.

### DIFF
--- a/etc/libinput/g502-wheelscroll-fix.quirks
+++ b/etc/libinput/g502-wheelscroll-fix.quirks
@@ -1,0 +1,3 @@
+[Logitech G502]
+MatchName=*Logitech G502*
+AttrEventCode=-REL_WHEEL_HI_RES;-REL_WHEEL_HI_RES;


raphipod:
It probably won't because it's more of a hack rather than a solution.

https://bugzilla.kernel.org/show_bug.cgi?id=216885 - that's the discussion regarding this issue.

Seemingly high-res scrolling is broken in libinput with Logitech mice since kernel 6.1.2

This is for the G903, although the [workaround](https://bugzilla.kernel.org/show_bug.cgi?id=216885#:~:text=bugs%20all%20related%3F-,Comment%2054,-David%20Roth%202024) at the end of the discussion also works with my wireless G502.

Some also suggested that blacklisting the two kernel modules 'hid_logitech_dj' and 'hid_logitech_hidpp' fixes the issue, but it prevents reporting of the battery charge on the desktop - so it's not a viable solution either.

Maybe I should add a page to the wiki with that fix rather than straight up adding something not everybody is affected by.

raphipod:
I will try to open a ticket in the libinput GitLab repo and then add the appropriate quirk. I didn't know that quirks are actually a part of libinput and encouraged. I thought they were merely a hacky workaround kind of thing. Thanks nonetheless, I'm sorry for any inconvenience caused.

1Naim:
> It seems that high-res scrolling is broken in libinput with Logitech mice since kernel 6.1.2, was fixed in 6.2.10, and re-appeared in 6.3.2/6.3.3

Have you tried testing these older kernels and go further as to bisect the commit that caused the regression? I think theres value in doing that, since theres precedent in it being fixed before.

PR #207, opened 2026-02-13
regdomain: Automatically set the wireless regularity domain

ptr1337:
Mostly taken from https://invent.kde.org/kde-linux/kde-linux/-/commit/8cc8ad3989d338eedae01d037728b358f0ebc69a

### DIFF
--- a/usr/lib/iw-set-regdomain
+++ b/usr/lib/iw-set-regdomain
@@ -1,0 +1,83 @@
+#!/bin/sh
+
+# SPDX-License-Identifier: ISC
+# SPDX-FileCopyrightText: 2025 Thomas Duckworth <tduck@filotimoproject.org>
+# SPDX-FileCopyrightText: 2009-2014 Red Hat, Inc.
+# SPDX-FileCopyrightText: 2025 Velocity Limitless, LLC.
+
+# See https://gitlab.com/VelocityLimitless/Projects/iw-setregdomain
+
+# Copyright 2009-2014 Red Hat, Inc.  All rights reserved.
+#
+# Permission to use, copy, modify, and/or distribute this software for any
+# purpose with or without fee is hereby granted, provided that the above
+# copyright notice and this permission notice appear in all copies.
+#
+# THE SOFTWARE IS PROVIDED "AS IS" AND THE AUTHOR DISCLAIMS ALL WARRANTIES
+# WITH REGARD TO THIS SOFTWARE INCLUDING ALL IMPLIED WARRANTIES OF
+# MERCHANTABILITY AND FITNESS. IN NO EVENT SHALL THE AUTHOR BE LIABLE FOR
+# ANY SPECIAL, DIRECT, INDIRECT, OR CONSEQUENTIAL DAMAGES OR ANY DAMAGES
+# WHATSOEVER RESULTING FROM LOSS OF USE, DATA OR PROFITS, WHETHER IN AN
+# ACTION OF CONTRACT, NEGLIGENCE OR OTHER TORTIOUS ACTION, ARISING OUT OF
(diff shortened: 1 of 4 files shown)


usr/lib/iw-set-regdomain
```diff
@@ -0,0 +1,83 @@
...
+	if [ -n "$COUNTRY" ]
+	then
+		/usr/bin/iw reg set $COUNTRY
+		exit
+	fi
+fi
+
+if [ -f "$LOCALTIME" ]
+then
+	ZONE=$(readlink -f $LOCALTIME)
```
ventureoo: Why not just query it via ``timedatectl show -P "Timezone"``?

ptr1337: This has been mainly picked from the KDE distribution. We can of course make changes

ptr1337:
> > No, because the device itself can also set specific frequencies. The device also needs to enable/disable these specific frequencies if not allowed. You can research in regulations.
> 
> My understanding is that this responsibility lies with the AP. A client can request but does not control spectrum usage. It is the AP that would be at fault for transmitting on an illegal band.

No, thats not true :)
https://wireless.docs.kernel.org/en/latest/en/developers/regulatory.html
https://www.wwt.com/blog/automated-frequency-coordination-afc-for-6ghz-wi-fi
https://approve-it.net/professional-blog/how-is-wi-fi-6-wi-fi-6e-and-wi-fi-7-impacting-global-regulatory-compliance/

I work in this sector for UWB in my daytoday job and its required to implement the detection on the client side. as example in russia its not allowed to enable UWB, either on client or AP/receiver side.

That is because you could interfere with signals from the military or other restricted frequency/usecases.

Right now on CachyOS to get Wifi 7 the user needs to set the US as region for example manually. This PR will set it automatically.

sirlucjan (approved):
```
lucjan at cachyos ~ 21:24:54    
❯ systemctl status cachyos-iw-set-regdomain.service
○ cachyos-iw-set-regdomain.service - Set Wireless Regulatory Domain on Timezone Change
     Loaded: loaded (/usr/lib/systemd/system/cachyos-iw-set-regdomain.service; static)
     Active: inactive (dead) since Fri 2026-02-27 21:21:03 CET; 3min 54s ago
 Invocation: f305361cb1fc444ab7734dd9ad63b2e3
    Process: 999 ExecStart=/usr/lib/iw-set-regdomain (code=exited, status=0/SUCCESS)
   Main PID: 999 (code=exited, status=0/SUCCESS)
   Mem peak: 2.5M
        CPU: 9ms

lut 27 21:21:03 cachyos systemd[1]: Starting Set Wireless Regulatory Domain on Timezone Change...
lut 27 21:21:03 cachyos systemd[1]: cachyos-iw-set-regdomain.service: Deactivated successfully.
lut 27 21:21:03 cachyos systemd[1]: Finished Set Wireless Regulatory Domain on Timezone Change.
lucjan at cachyos ~ 21:24:57    
❯ iw reg get                                       
global
country PL: DFS-ETSI
        (2400 - 2483 @ 40), (N/A, 20), (N/A)
        (5150 - 5250 @ 80), (N/A, 23), (N/A), NO-OUTDOOR, AUTO-BW
        (5250 - 5350 @ 80), (N/A, 20), (0 ms), NO-OUTDOOR, DFS, AUTO-BW
        (5470 - 5725 @ 160), (N/A, 26), (0 ms), DFS
        (5725 - 5875 @ 80), (N/A, 13), (N/A)
        (5945 - 6425 @ 320), (N/A, 23), (N/A), NO-OUTDOOR
        (57000 - 66000 @ 2160), (N/A, 40), (N/A)
```

Barmem:
Can I disable it?

ptr1337:
> Can I disable it?

Yes, you can remove iw or mask the user service

PR #211, opened 2026-02-25
Refactor I/O udev rules for stability and robustness

damachine:
Suggestion:
I know this is currently the arch way to set the udev rules for the I/O scheduler. Maybe they could be optimized for greater stability and robustness. Currently, they rely solely on kernel naming conventions, which is acceptable but can be somewhat imprecise.

I propose adding explicit subsystem filtering and excluding virtual devices such as loopbacks or zram. This will keep the logic clearer and make the rules more future-proof should kernel patterns be expanded in the future.

Note: Although `KERNEL!="loop*"` and `KERNEL!="zram*"` are technically redundant given the current specific kernel matches (`sd[a-z]*`, `mmcblk*`, `nvme*`), they are retained for clarity and robustness. This prevents accidental application to virtual devices and in case if the positive match patterns should be expanded in the future.

### DIFF
--- a/usr/lib/udev/rules.d/60-ioschedulers.rules
+++ b/usr/lib/udev/rules.d/60-ioschedulers.rules
@@ -4,7 +4,7 @@ ACTION=="add|change", KERNEL=="sd[a-z]*", ATTR{queue/rotational}=="1", \
 
 # SSD
 ACTION=="add|change", KERNEL=="sd[a-z]*|mmcblk[0-9]*", ATTR{queue/rotational}=="0", \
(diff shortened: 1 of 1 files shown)


ventureoo (changes requested):
I don't see why all of these  `KERNEL!=` checks are needed if it's already explicitly set in `KERNEL==`

damachine:
Thanks for the review! Sorry, I wasn't aware that support for rotating NVMe SSDs already existed. https://lore.kernel.org/linux-block/968b578f-146c-4e94-a68e-e7451762a81a@kernel.dk/

PR #220, opened 2026-03-13
ioschedulers: switch NVMe scheduler from none to kyber

ptr1337:
While none provides marginally better throughput on empty or lightly used NVMe drives, performance degrades under mixed workloads as the drive fills up. Kyber is designed for fast multi-queue devices and enforces latency targets by throttling queue depth under congestion, providing more consistent read/write latency on drives that are 50%+ filled, which reflects real-world desktop usage.

### DIFF
--- a/usr/lib/udev/rules.d/60-ioschedulers.rules
+++ b/usr/lib/udev/rules.d/60-ioschedulers.rules
@@ -8,4 +8,4 @@ ACTION=="add|change", KERNEL=="sd[a-z]*|mmcblk[0-9]*", ATTR{queue/rotational}=="
 
 # NVMe SSD
 ACTION=="add|change", KERNEL=="nvme[0-9]*", ATTR{queue/rotational}=="0", \
-    ATTR{queue/scheduler}="none"
+    ATTR{queue/scheduler}="kyber"


ptr1337:
I think this would make sense doing so, since the therotical throughput would be better with None, but as soon the Disk gets more filled this gets worse.

In real life scenarious there is barely any disk not up to 50% filled, also it would help mitigating stutters with kyber.

I search the recent study about this and post the link

ventureoo (approved):
I still have some doubts about that, mainly because I didn't notice much difference during testing on my rig under fio stress test, but due to comments in https://github.com/CachyOS/CachyOS-Settings/pull/42#issuecomment-1902184675 I think we can try it again.

ptr1337:
Lets try

ghost:
> I search the recent study about this and post the link

Could you share the link with us? I can't seem to find the post in the CachyOS forums or wherever it's supposed to be.

Everywhere I look, I only see "none" listed as the recommendation and currently the best option for NVMe. That's why I'd like to take a look at the study.

ptr1337:
> > I search the recent study about this and post the link
> 
> Could you share the link with us? I can't seem to find the post in the CachyOS forums or wherever it's supposed to be.
> 
> Everywhere I look, I only see "none" listed as the recommendation and currently the best option for NVMe. That's why I'd like to take a look at the study.

https://research.spec.org/icpe_proceedings/2024/proceedings/p154.pdf

There you go.

PR #224, opened 2026-04-29
net: disable WiFi power save by default

Sergio-N-Jr:
Adds a NetworkManager configuration to disable WiFi power saving by default.
                                                                                                        
Power save mode introduces latency spikes and inconsistent throughput,
which conflicts with CachyOS's focus on responsiveness. This is a common                              
tweak recommended for gaming and low-latency workloads.                                             
                                                                                                        
The setting uses NetworkManager's native configuration instead of iw                                  
commands, so it works regardless of the interface name (wlan0, wlp3s0, etc.)                          
and is applied automatically on reconnect.                                                            
                                                                                                        
wifi.powersave = 2 → disabled                                                                         
wifi.powersave = 3 → enabled (NM default)

### DIFF
--- a/usr/lib/NetworkManager/conf.d/wifi-powersave.conf
+++ b/usr/lib/NetworkManager/conf.d/wifi-powersave.conf
@@ -1,0 +1,2 @@
+[connection]
+wifi.powersave = 2


1Naim:
> Power save mode introduces latency spikes and inconsistent throughput

How bad is it? Do you have some results?

Sergio-N-Jr:
Yes, it's quite noticeable.

With power save enabled, my ping to the local router (`192.168.x.x`) was consistently spiking even on a local network with no internet involved.

After applying `wifi.powersave = 2`, the latency dropped significantly and became stable.

This is especially relevant for gaming and any low-latency workload, which aligns well with CachyOS's performance focus.

1Naim:
I'm asking for the numbers here.

Sergio-N-Jr:
Yes!

Here are some quick numbers from my local router ping test (`192.168.0.1`).

Power Save ON
Samples: `250, 270, 289, 309, 329, 351, 374, 387, 408 ms`  
Average: **329.7 ms**

Power Save OFF
Samples: `134, 153, 172, 188, 210 ms`  
Average: **171.4 ms**

Difference
Disabling power save reduced average latency by  ~158 ms (about 48% lower).

Revival8697:
> Power Save OFF
> Samples: `134, 153, 172, 188, 210 ms`
> Average: **171.4 ms**

So `min/avg/max/mdev = 134/171.4/210/26.46 ms`. That is bad.

> Disabling power save reduced average latency by ~158 ms (about 48% lower).

It is still bad, you should fix your network. Here is mine:

- Power save on: `rtt min/avg/max/mdev = 9.866/11.161/15.150/1.544 ms`
- Power save off: `rtt min/avg/max/mdev = 1.979/2.938/7.430/1.629 ms`

Sergio-N-Jr:
Yes, you're right — the latency is still bad in that state, and my network setup clearly needs further tuning.

I found that the following command fixes the issue completely on my side:

```bash
sudo iw dev wlan0 set txpower fixed 2000
```

After applying it, my average ping drops to around 5–7 ms, making the connection stable again.

So in my case:

Power save OFF already reduces latency significantly.
TX power adjustment removes the remaining problem.

However, this commit is only about disabling power save, since that is a much safer and more universal default.

Changing TX power may involve regulatory restrictions depending on the country, device, or driver support, so I still need to research it properly before recommending it as a system default.

Once I finish testing and studying it further, I may open another PR with a possible complete solution.

Sergio-N-Jr:
Anyway, for now, temporarily setting **Power Save OFF** by default would likely help many users.

mati865:
I had similar power saving issues with Qualcomm NCM865 (numbers in the name are just a coincidence) last year. Wi-Fi would struggle to achieve 100 Mb/s and the latency was jittery. Eventually it got fixed in the kernel/firmware and works fine now:
```
❯ trip -m pretty 192.168.50.1
┌─────┬──────────────┬─────────────────────┬───────┬─────┬──────┬──────┬─────┬──────┬──────┬────────┐
│ Hop ┆ IPs          ┆ Addrs               ┆ Loss% ┆ Snt ┆ Recv ┆ Last ┆ Avg ┆ Best ┆ Wrst ┆ StdDev │
╞═════╪══════════════╪═════════════════════╪═══════╪═════╪══════╪══════╪═════╪══════╪══════╪════════╡
│ 1   ┆ 192.168.50.1 ┆ RT-AX59U-BA32.local ┆ 0.0   ┆ 10  ┆ 10   ┆ 3.4  ┆ 2.9 ┆ 2.2  ┆ 4.1  ┆ 0.5    │
└─────┴──────────────┴─────────────────────┴───────┴─────┴──────┴──────┴─────┴──────┴──────┴────────┘
```
This is over 5 GHz at -64 dBm signal level with power save on.

I recommend monitoring how your adapter behaves during normal desktop usage and during stress scenarios like speedtest.net with `wavemon` tool and reporting the issue with the findings at the kernel bug tracker and to your adapter manufacturer.

ventureoo:
I would prefer if it possible to go away from NetworkManager configuration and use udev rule for that.

1Naim:
> I would prefer if it possible to go away from NetworkManager configuration and use udev rule for that.

I agree, it should enable powersaving (or whatever the default is) when running on battery.

PR #239, opened 2026-06-25
modprobe: enable nvidia_drm modeset and fbdev by default

romale23:
Without nvidia_drm modeset=1 and fbdev=1 set, NVIDIA users on Wayland
experience PAT write-combining conflicts that cause Xid 31 MMU faults,
particularly when browsers with GPU acceleration (Brave, Chrome) allocate
GPU textures.

Kernel log showing the issue:
x86/PAT: conflicting memory types uncached-minus <-> write-combining
ioremap memtype_reserve failed -16
NVRM: Xid 31 MMU Fault: ENGINE GRAPHICS GPC0 GPCCLIENT_RAST

These options are required for proper KMS operation on Wayland with NVIDIA
drivers 550+ and should be set by default for all CachyOS NVIDIA users.


damachine:
Hey, any specific reason for this? It's already been enabled since upstream 580+, so this should be redundant.

romale23:
The crash occurred on a fresh 580 install without these explicitly set. Can you confirm the driver enables modeset and fbdev by default unconditionally? If so I'll close this and wonder why i'm special

damachine:
> The crash occurred on a fresh 580 install without these explicitly set. Can you confirm the driver enables modeset and fbdev by default unconditionally? If so I'll close this and wonder why i'm special

Normally it should already be active. You can check it by running:

`cat /sys/module/nvidia_drm/parameters/modeset`

and

`cat /sys/module/nvidia_drm/parameters/fbdev`

Both should return `Y` if it's enabled. So if it active without your hack the error must be a different

romale23:
I manually enabled it today after i found it wasn't there for me. I'm guessing that means as usually something is just wrong for me. I'll close the PR sorry to bother y'all

PR #241, opened 2026-06-29
Enable REISUB by default

phoropter:
Add ability to perform REISUB by default


etc/sysctl.d/99-sysrq.conf
```diff
@@ -0,0 +1 @@
+kernel.sysrq = 1
```
ptr1337, line 1: Mhm ,cant we put into cachyos sysctl?

phoropter: Yea definitely could, I didn't think about that.

ventureoo (approved):

Issue #191, opened 2025-12-06
Watchdog (wdat_wdt) not disabled by default on CachyOS


inkatail:
While [it's fixed in the handheld version](https://github.com/CachyOS/CachyOS-Handheld/issues/37), no one ever really bothered blacklisting it in the desktop version, maybe because most desktops don't have the `wdat_wdt` module being enabled.
Sadly or not, I am an owner of a bad laptop which definitely does enable it (I checked in `wdctl`) 

I don't know why `nowatchdog` doesn't disable it in first place but yeah.

ventureoo:
Fixed by: https://github.com/CachyOS/CachyOS-Settings/commit/6053ac9f624924951269d64153b14e6eea18b30f

PR #242, opened 2026-07-01
Blacklists wdat_wdt as well to address #191

ThomasADarnell:
nowatchdog is still being passed, so this just adds the blacklist that is also on handheld as suggested by #191


ventureoo:
I wonder if `sysctl -w kernel.nmi_watchdog=0` or `sysctl -w kernel.watchdog=0` actually helps with that too. Can someone check it?

ventureoo (approved):

PR #243, opened 2026-07-07
cachyos-settings: add nice value to limits.d

Jinra:
Current settings will see this error with the user pipewire service due to missing permissions. This PR adds the nice value that pipewire is trying to set to limits.d.

image

Adding the value solves the errors
image


ptr1337 (approved):

1Naim:
isn't this intended? ur supposed to install `realtime-privileges` and add your user to the realtime group to make that work.

ptr1337:
> isn't this intended? ur supposed to install `realtime-privileges` and add your user to the realtime group to make that work.

Maybe we should do this then in the installer

PR #245, opened 2026-07-16
20-audio-pm.rules: fix filtering Mains, empty state file, hardcoded BAT0

droserasprout:
Hi! I brought you some bugfixes. Skipping the Issue step because these changes do not change behavior or settings, hope it's okay.

## Problem

Three defects in the `snd_hda_intel` power-management rules, confirmed on a live CachyOS laptop (ASUS G14 GA401IV, AC0 + BAT0 + a USB-C source PSY):

1. **USB-C ports re-enable codec power saving on AC.** Rules 2 and 3 matched any `power_supply` uevent with `POWER_SUPPLY_ONLINE=0|1` and no type filter. A USB-C port PSY reports `ONLINE=0` and fires a uevent on every plug/unplug, each writing `power_save=10` *while on AC mains* — reintroducing the audio crackle this file exists to prevent.

2. **Empty state file when booting on battery.** `touch` was unconditional but the value was only written on AC, so a battery boot left the file empty. The `cat file || echo 10` fallback was dead code (`cat` on an empty file succeeds), writing a bare newline that sysfs rejected with EINVAL. The intended default of 10 never applied.

3. **`BAT0` hardcoded.** Machines exposing only `BAT1` (some ThinkPads, Framework, Surface) read empty — `!= "Discharging"` — so they were treated as AC-powered and lost power saving on battery.

## Fix

- Add `ENV{POWER_SUPPLY_TYPE}=="Mains"` to both `power_supply` rules so only the AC adapter toggles power saving.
- Drop the unconditional `touch`; write the saved value only on AC and only when non-zero, and have the restore path test for a non-empty value. Absent/empty state now falls back to 10.
- Replace hardcoded `BAT0` with a glob over `/sys/class/power_supply/BAT*/status`, so any battery name works and a batteryless desktop is treated as always-on-AC.

## Testing

- `udevadm verify` passes.
- On a live laptop: `power_save` stays `0` across USB-C plug/unplug on AC, flips to `10` on unplugging mains, returns to `0` on reconnect. Battery-boot path no longer logs an EINVAL error.

### DIFF
--- a/usr/lib/audio-pm-ctl
+++ b/usr/lib/audio-pm-ctl
@@ -1,0 +1,67 @@
+#!/usr/bin/bash
+
+# Disables power saving capabilities for `snd-hda-intel` when device is connected to AC.
+# Prevents audio cracks on some hardware. The saved value is stashed in /run and restored
+# when the machine switches back to battery.
+#
+# Invoked from `20-audio-pm.rules` with one of: init, battery, ac.
+
+PARAM=/sys/module/snd_hda_intel/parameters/power_save
+STATE=/run/udev/snd-hda-intel-powersave
+
+# Stash the current `power_save` value to restore it later.
+save() {
+    local cur
+    cur=$(cat "$PARAM")
+    [[ $cur != 0 ]] && echo "$cur" > "$STATE"
+    return 0
+}
+
+# True if at least one mains supply is currently online.
+is_on_ac() {
+    local d type online
+    for d in /sys/class/power_supply/*; do
+        type=$(cat "$d/type" 2>/dev/null)
+        [[ $type == Mains ]] || continue
+        online=$(cat "$d/online" 2>/dev/null)
+        [[ $online == 1 ]] && return 0
+    done
+    return 1
+}
+
+# Sound card appeared: disable power saving when on AC.
+do_init() {
+    is_on_ac || return 0
+    save
+    echo 0 > "$PARAM"
+}
+
+# Switched to battery: restore the saved value (default 10).
+do_battery() {
+    local v
+    v=$(cat "$STATE" 2>/dev/null)
+    [[ -n $v ]] || v=10
+    echo "$v" > "$PARAM"
+}
+
+# Switched to AC: disable power saving.
+do_ac() {
+    save
+    echo 0 > "$PARAM"
+}
+
+usage() {
+    echo "usage: $0 {init|battery|ac}" >&2
+    exit 1
+}
+
+main() {
+    case "$1" in
+        init)    do_init ;;
+        battery) do_battery ;;
+        ac)      do_ac ;;
+        *)       usage ;;
+    esac
+}
+
+main "$@"
--- a/usr/lib/udev/rules.d/20-audio-pm.rules
+++ b/usr/lib/udev/rules.d/20-audio-pm.rules
@@ -1,17 +1,12 @@
-# Disables power saving capabilities for snd-hda-intel when device is not
-# running on battery power. This is needed because it prevents audio cracks on
-# some hardware.
+# Disables power saving capabilities for `snd-hda-intel` when device is connected to AC.
+# Prevents audio cracks on some hardware. The saved value is stashed in /run and restored
+# when the machine switches back to battery. See `usr/lib/audio-pm-ctl`.
+
 ACTION=="add", SUBSYSTEM=="sound", KERNEL=="card*", DRIVERS=="snd_hda_intel", TEST!="/run/udev/snd-hda-intel-powersave", \
-    RUN+="/usr/bin/bash -c 'touch /run/udev/snd-hda-intel-powersave; \
-        [[ $$(cat /sys/class/power_supply/BAT0/status 2>/dev/null) != \"Discharging\" ]] && \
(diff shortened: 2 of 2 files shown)


ventureoo:
Changes themselves are fine, but I think it would be better if we move all inlined shell code to a seperate script, because reading all of this is actually painful. You can call it `audio-pm-ctl` or something like that and take action as first argument.

usr/lib/udev/rules.d/20-audio-pm.rules
```diff
@@ -1,17 +1,18 @@
...
 # running on battery power. This is needed because it prevents audio cracks on
-# some hardware.
+# some hardware. The saved value is stashed in /run so it can be restored when
+# the machine switches back to battery.
 ACTION=="add", SUBSYSTEM=="sound", KERNEL=="card*", DRIVERS=="snd_hda_intel", TEST!="/run/udev/snd-hda-intel-powersave", \
-    RUN+="/usr/bin/bash -c 'touch /run/udev/snd-hda-intel-powersave; \
-        [[ $$(cat /sys/class/power_supply/BAT0/status 2>/dev/null) != \"Discharging\" ]] && \
-        echo $$(cat /sys/module/snd_hda_intel/parameters/power_save) > /run/udev/snd-hda-intel-powersave && \
-        echo 0 > /sys/module/snd_hda_intel/parameters/power_save'"
+    RUN+="/usr/bin/bash -c 'grep -qx Discharging /sys/class/power_supply/BAT*/status 2>/dev/null || { \
```
ventureoo: So, where is the guarantee that all these BATs* will actually be main batteries and not related to other devices?

usr/lib/udev/rules.d/20-audio-pm.rules
```diff
@@ -1,17 +1,18 @@
...
+        cur=$$(cat /sys/module/snd_hda_intel/parameters/power_save); \
+        [[ $$cur != 0 ]] && echo $$cur > /run/udev/snd-hda-intel-powersave; \
+        echo 0 > /sys/module/snd_hda_intel/parameters/power_save; }'"
 
-SUBSYSTEM=="power_supply", ENV{POWER_SUPPLY_ONLINE}=="0", TEST=="/sys/module/snd_hda_intel", \
-    RUN+="/usr/bin/bash -c 'echo $$(cat /run/udev/snd-hda-intel-powersave 2>/dev/null || \
-        echo 10) > /sys/module/snd_hda_intel/parameters/power_save'"
+ACTION=="add|change", SUBSYSTEM=="power_supply", ENV{POWER_SUPPLY_TYPE}=="Mains", ENV{POWER_SUPPLY_ONLINE}=="0", TEST=="/sys/module/snd_hda_intel", \
```
ventureoo, line 8: I'm also not sure if `ACTION=="add"` is the right event here.

usr/lib/udev/rules.d/20-audio-pm.rules
droserasprout: Yep, the correct way is to check AC+Mains, the same way udev rules match. Fixed.

But out of curiosity... I haven't seen devices/drivers on x86-64 that put something other than ACPI batteries in `/sys/class/power_supply/BAT*` yet. Could you provide an example?

usr/lib/udev/rules.d/20-audio-pm.rules
droserasprout: `systemd-udev-trigger.service` calls `udevadm trigger --type=all --action=add` to coldplug devices, so I think yes, it should be included to apply PM settings on boot.

ventureoo (changes requested):
Could you please take each action handler in a particular function and use local variables for it?

usr/lib/audio-pm-ctl
```diff
@@ -0,0 +1,67 @@
+#!/usr/bin/bash
```
ventureoo, line 1: I think we should put that script rather in ``/usr/bin``.

droserasprout: I would leave it in `/usr/lib` for internal use only, since a user who discovered this script and invoked it manually would be fighting with udev.

droserasprout:
Sorry, I won't go into another round of tuning this ugly script to perfection, primarily because I have removed it completely from my system after finding out who's messing with sound and opening this PR 😆

I couldn't find reliable data on how audio powersave affects real power consumption, and I haven't observed any difference personally. But audio glitches are very real on my hardware.

Have you considered simply dropping this hack?

usr/lib/audio-pm-ctl
ventureoo: Then it should be placed in `/usr/lib/udev` by convention.

ventureoo:
> I couldn't find reliable data on how audio powersave affects real power consumption, and I haven't observed any difference personally. But audio glitches are very real on my hardware.

The reason why we add these rules was exactly because we wanted to fix some audio crackling issues, enabling the power saving only when working from a battery. Do you want to say that using these rules also leads to audio cracks?

ventureoo:
@droserasprout Any progress? I actually would like to merge it if you move `audioctl` script to `/usr/lib/udev` directory.

ventureoo (approved):

PR #246, opened 2026-07-16
iw-set-regdomain: broaden no-country zones, resolve aliases

droserasprout:
Fixes two papercuts in the timezone to regulatory-domain lookup.

**Broaden the no-country whitelist**

The old `case` block matched only 10 UTC/GMT aliases, so 46 other country-less zones fell through to `logger -s` + `exit 1`. They now exit 0 quietly.

**Resolve backward-compat aliases**

`zone.tab` only lists canonical `Region/City` names, so a user on a legacy alias like `Asia/Calcutta` got "Could not determine country" and no regdomain. Aliases are now canonicalized via tzdata's Link table (`/usr/share/zoneinfo/tzdata.zi`) before the lookup. The no-country whitelist is checked first, so generic zones like `CET` and `EST` still skip rather than being pinned to an arbitrary single country.

**Testing**

Swept all 598 `timedatectl list-timezones` entries against the real script:

- 542 resolve a country from `zone.tab`  (418 canonical + 124 via alias)
- 56 hit the graceful no-country exit
- 0 unhandled

### DIFF
--- a/usr/lib/iw-set-regdomain
+++ b/usr/lib/iw-set-regdomain
@@ -55,19 +55,30 @@ then
 	exit 1
 fi
 
+# Zones with no single associated country
+case "$ZONE" in
+	UTC|UCT|Universal|Zulu|GMT|GMT[+-]0|GMT0|Greenwich|Factory|posixrules|\
+	CET|EET|WET|MET|EST|MST|HST|EST5EDT|CST6CDT|MST7MDT|PST8PDT|\
+	Etc/*)
+		$LOGGER "Timezone is $ZONE, with no associated country. Not setting wireless regulatory domain."
+		exit 0
+		;;
+esac
+
+# Resolve backward-compat aliases (e.g. Asia/Calcutta) to their canonical name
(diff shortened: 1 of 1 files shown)


ptr1337:
• zone.tab: 418 zones — zone1970.tab: 312 zones
   • Zones only in zone1970.tab (gained): 0
   • Zones only in zone.tab : 106

droserasprout:
Oh, sorry, I have become muddled by these two files. Gonna check everything more carefully. 

UPD: Correct way to resolve aliases was `tzdata.zi` lookup. Updated code and description.

usr/lib/iw-set-regdomain
```diff
@@ -26,6 +26,12 @@ REGDOMAIN=/etc/iw-regdomain
 
 LOGGER="/usr/bin/logger -t iw-set-regdomain"
 
+if [ ! -x /usr/bin/iw ]
```
ventureoo: This looks like useless check, since we depend on on `iw` anyway.

vnepogodin (approved):

ventureoo (approved):

Issue #250, opened 2026-08-08
Improving cachyos-bugreport.sh redaction


drinkingoutofcups:
Piggybacking off #227, I think some improvements can be made to the redaction system.

From reading the bug report on my own PC, it doesn't seem like anything personally identifying will ever be leaked in the first few sections, so running sed on those lines only adds a risk of giving away information through redaction. The dmesg and journalctl are the sections that may reveal identifying information. I can see some users in the forum have "cachyos" as their hostname because it is redacted in this section. 
Although I just said we should not redact the first few sections, we may want to redact UUIDs in all sections. Machine UUID and storage UUIDs are found in top sections and in dmesg from /dev/disk/by-uuid/ messages.
I had over 1000 files leaked by xdg-desktop-portal-kde:
`Aug 07 12:05:47 <hostname-redacted> xdg-desktop-portal-kde[1474]: No node found for item that was just removed: QUrl("file:///mnt/drive/folder/file.txt")`
We could filter xdg-desktop-portal-kde or redact anything from `"file://` to the ending quotation mark. I don't think either of these are likely to be relevant to the subject of a user's bug report.
I also noticed UFW shows IPv6 addresses unredacted.

There's a lot of required information that is personally identifying in the bug report so this may not be that helpful, but I think the principle stands that we shouldn't require any unnecessary information. If anyone else thinks this is worth doing I can work on this.

petejm:
_Disclosure: posted by an AI agent (Claude) on behalf of @petejm, who reviewed and approved this comment._

Agreed, and the hostname case is worse than an oracle: it destroys the data a maintainer opens the report for.

The default CachyOS hostname is generic, so `hostname == cachyos` is common. I reproduced `redact()`'s sed pipeline against a synthetic log (no real data). On such a machine:

```
input:    uname: Linux cachyos 7.1.6-1-cachyos
          cachyos-v4/linux-cachyos 7.1.6-1 [installed]

output:   uname: Linux <hostname-redacted> 7.1.6-1-<hostname-redacted>
          <hostname-redacted>-v4/linux-<hostname-redacted> 7.1.6-1 [installed]
```

The package and repo names are gone, and the mangling pattern announces the hostname anyway.

[...]

PR #252, opened 2026-08-14
Improved bugreport redaction

drinkingoutofcups:
From the discussion in #250, I added new matches for SSIDs, IPv6, UUID, "file://", and a condition to not remove hostname when it's "cachyos". Also made some of the printout consistent.

### DIFF
--- a/usr/bin/cachyos-bugreport.sh
+++ b/usr/bin/cachyos-bugreport.sh
@@ -67,32 +67,27 @@ bugreport() {
 
     cat << EOF >"$LOG_FILENAME"
 ____________________________________________
-
 Start of CachyOS bug report log file. Please send this report,
 along with a description of your bug, to CachyOS.
 
 Date: $(date)
 uname: $(uname -a)
 cmdline: $(cat /proc/cmdline)
-
 ____________________________________________
-Getting Hardware Information
+Getting hardware information
 
 $(inxi -Farz)
 ____________________________________________
-Getting Scheduler information
+Getting scheduler information
 
 sched-ext:
 $(grep -R "" /sys/kernel/sched_ext/ 2>/dev/null || echo "sched_ext not available")
 
 $(journalctl --output cat -k | grep -i scheduler || true)
-
 ____________________________________________
-
 dmesg
 
 $(dmesg)
-
 ____________________________________________
 journalctl of current boot
 
@@ -102,11 +97,10 @@ journalctl of previous boot
 
 $(journalctl -b -1 -p 4..1 2>/dev/null || echo "No previous boot log available")
 ____________________________________________
-
 Installed packages
 
 $(get_installed_packages)
---------------------------------------------
+____________________________________________
 EOF
 }
 
@@ -121,7 +115,9 @@ redact() {
     # Redact hostname (appears in uname, dmesg, journalctl)
     local hn
     hn=$(hostname)
-    sed_args+=(-e "s|$(sed_escape "$hn")|<hostname-redacted>|g")
+    if [[ $hn != "cachyos" ]]; then
+        sed_args+=(-e "s|$(sed_escape "$hn")|<hostname-redacted>|g")
+    fi
 
     # Redact real username and home directory (SUDO_USER is set when run via sudo)
     local real_user="${SUDO_USER:-}"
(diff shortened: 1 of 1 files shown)


petejm:
Thanks for picking this up, and for the offer to test. I ran the PR head against a synthetic
report, and the real script with stubbed collectors, on GNU bash 5.3.15 and GNU sed 4.10. Two
of these fail on ordinary machines, so I have put those first.

## 1. An empty SSID list produces an empty sed expression

`nmcli -t -f name,type c | sed -nE 's/(.*)\:.*wireless/\1/p'` outputs nothing on a machine
with no saved wireless profile. The `<<<` herestring still yields one iteration, so the loop
appends an empty regex:

```
[-e]
[s||<ssid-redacted>|g]
```

What happens next depends on whether an earlier `-e` was added:

* If the hostname guard skipped and `SUDO_USER` is unset, this is the first expression:
  `sed: -e expression #7, char 0: no previous regular expression`, exit 1. Under
  `set -euo pipefail` the script stops before `upload`, which is the safe direction, but it
  leaves `cachyos-bugreport.log` on disk completely unredacted, and that is the filename
  users are told to attach.
* If any earlier `-e` exists, GNU sed reuses the last executed regex, so the SSID pass
  silently does nothing. Exit 0, no warning.

I hit the same abort two other ways, both verified: `nmcli` not installed (exit 127, for
anyone on systemd-networkd or iwd) and NetworkManager installed but not running (exit 8).
Both leave the unredacted log.

Suggested gate: `[[ -n $ssid_list ]]` before the loop, `[[ -n $ssid ]] || continue` inside it,
and `command -v nmcli` or a trailing `|| true` on the collection so an absent NetworkManager
cannot abort the whole redaction pass.

## 2. The hostname guard misses three of the four hostnames CachyOS ships

`[[ $hn != "cachyos" ]]` is an exact, case-sensitive comparison. It covers the CLI installer
default and the kernel fallback, but not the other three:

| Shipped value | Source | Guard |
|---|---|---|
| `cachyos` | `linux-cachyos/config:66` `CONFIG_DEFAULT_HOSTNAME`, and the `New-Cli-Installer` default | matches |
| `cachyos-${cpu}` | `calamares-config`, `etc/calamares/modules/users.conf:77` `template:` | misses |
| `CachyOS` | `CachyOS-Live-ISO` and `CachyOS-CLI-ISO`, `archiso/airootfs/etc/hostname` | misses |
| `archlinux` | `linux-cachyos-hardened/config:64` | misses |

The calamares row is the graphical installer, so a normal desktop install produces something
like `cachyos-znver4` and gets redacted despite the guard.

The ISO row is the one I would prioritise. Running the PR head with hostname `CachyOS`
rewrites the report's own boilerplate:

```
Start of <hostname-redacted> bug report log file. Please send this report,
along with a description of your bug, to <hostname-redacted>.
```

together with every `Distro: CachyOS` line from inxi. A report generated on the live ISO stops
saying which distribution it is a report for.

A case-insensitive prefix test covers all four. Either `shopt -s nocasematch` with
`[[ $hn != cachyos* ]]`, or `[[ ${hn,,} != cachyos* ]]` if bash 4 is acceptable here, plus
`archlinux` and the empty string.

The empty hostname matters on its own: the substitution becomes `s||<hostname-redacted>|g` in
first position and aborts the entire pass exactly as in point 1.

## 3. Three ways the SSID rule can miss silently

These concern me more than a loud failure, because the resulting log looks redacted.

* `nmcli -t` escapes `:` in names as `\:`. The capture keeps the backslash and `sed_escape`
  escapes it again, so a real SSID `Cafe:Wifi` is never matched.
* `-f name,type` returns the connection profile name, not the SSID. The two coincide by
  default and diverge on any renamed profile. The actual field is
  `nmcli -g 802-11-wireless.ssid c show <name>`; the listing form rejects
  `-f 802-11-wireless.ssid`, so it has to be queried per connection.
* The SSID substitution has no `\b` anchors, unlike the hostname rule above it. A short SSID
  substring-matches anything: with SSID `on`, `nvidia driver 570.86 loaded on card0` becomes
  `nvidia driver 570.86 loaded <ssid-redacted> card0`.

## Smaller notes

* The `file://` pattern is greedy. `.\+` runs to the last quote on the line, so
  `opened "file:///home/u/a.txt" then errno 5 on device "sda"` collapses to
  `opened "file://<file-redacted>"`.
* The UUID rule also scrubs `root=UUID=...` out of the kernel cmdline, which is usually
  load-bearing for boot and filesystem triage.
* The IPv6 pattern matches only the fully expanded eight-group form, as its comment says.
  Journal and dmesg output is nearly always `::` compressed.

Happy to send any of this as a patch on top of your branch if that is easier than reworking it.

_Disclosure: Researched and drafted by an AI agent (Claude), guided & reviewed by @petejm_

usr/bin/cachyos-bugreport.sh
```diff
@@ -121,7 +115,9 @@ redact() {
     # Redact hostname (appears in uname, dmesg, journalctl)
     local hn
     hn=$(hostname)
-    sed_args+=(-e "s|\b$(sed_escape "$hn")\b|<hostname-redacted>|g")
+    if [[ $hn != "cachyos" ]]; then
```
ventureoo, line 118:
```suggestion
    if [[ "$hn" != "cachyos" ]]; then
```

usr/bin/cachyos-bugreport.sh
```diff
@@ -121,7 +115,9 @@ redact() {
     # Redact hostname (appears in uname, dmesg, journalctl)
     local hn
     hn=$(hostname)
```
ventureoo, line 117:
```suggestion
    hn="$(hostname)"
```

ventureoo (changes requested):

petejm:
_Disclosure: posted by an AI agent (Claude) on behalf of @petejm, who reviewed and approved this comment._

Process note rather than a review point.

#255 is also open against the same redaction code in `usr/bin/cachyos-bugreport.sh`, and also
references #250. It takes a broader approach: enumerate the machine's own values first and
redact those, plus a test suite for the redaction path.

The two will conflict, since both edit the same expressions in the same file. I have reviewed
both and have no stake in which one lands. Flagging it so the choice gets made deliberately
rather than showing up later as a merge conflict.

Separately, on your note in #250 that "more testing would be nice": happy to run the same
reproduction set I used on #255 against this PR if that would be useful. cc @alexsmolya

PR #254, opened 2026-08-14
topmem: reject an invalid process count instead of ignoring it

NicoFilips:
tonumber() signals failure by returning nil rather than raising, so the pcall() wrapper always reported success and the error path was dead code: "topmem foo" fell through with size = nil and silently printed the default top 10. Check the converted value directly and also reject counts below 1, which previously printed just the header row.


usr/bin/topmem
```diff
@@ -203,9 +203,11 @@ local function main()
         elseif arg[i] == "--help" or arg[i] == "-h" then
             print_help()
         else
-            local status, value = pcall(tonumber, arg[i])
-            if not status then
-                die("The argument must be a number")
+            -- tonumber() reports failure by returning nil, it does not raise,
+            -- so pcall() always succeeded here and bad input was ignored
```
ventureoo, line 207: I think we can omit this comment, since `pcall` was removed here.

ventureoo:
Why this is marked as draft?

PR #255, opened 2026-08-14
cachyos-bugreport: redact enumerated machine values

alexsmolya:
## Summary

- enumerate current sensitive machine values and redact those exact literals instead of globally guessing at address-like strings
- suppress or redact hostnames only in structured report positions, preserving useful `linux-cachyos`, `cachyos-v4`, and kernel-version diagnostics
- retain narrow fallbacks for historical journal fields and add deterministic synthetic regression coverage

This addresses #250.

## Why this approach

Compared with #252, this avoids the broad IPv4 substitution that can turn a firmware version such as `6.18.44.1` into a false positive. It also handles privacy cases that implementation misses: compressed IPv6, 32-character machine IDs, USB serial numbers, nonstandard filesystem UUIDs, and empty Wi-Fi inventories.

Current values are collected from reliable system sources for IPv4/IPv6 addresses, MAC addresses, machine IDs, filesystem UUIDs, saved SSIDs, USB serials, and the invoking user's name/home. Historical previous-boot values that may no longer exist in current state are covered only in structured UFW, DHCP, SSID, UUID, machine-id, and USB fields.

Journal hostnames are suppressed at collection time, while the `uname` hostname is redacted positionally. Quoted `file://` paths are redacted one URL at a time so multiple URLs on a line do not consume useful intervening diagnostics.

## Tests

- `tests/test-bugreport-redaction.sh`
- `bash -n usr/bin/cachyos-bugreport.sh`
- `bash -n tests/test-bugreport-redaction.sh`
- `git diff HEAD^ --check`

The synthetic test covers current and historical IPv4/IPv6, MAC addresses, machine IDs, filesystem UUIDs, SSIDs, USB serials, username/home paths, multiple quoted file URLs, generic hostnames, preserved CachyOS package/kernel strings, the firmware-version false positive, and empty inventory behavior.

### DIFF
--- a/tests/test-bugreport-redaction.sh
+++ b/tests/test-bugreport-redaction.sh
@@ -1,0 +1,93 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+repo_root=$(cd -- "$(dirname -- "${BASH_SOURCE[0]}")/.." && pwd)
+# shellcheck source=../usr/bin/cachyos-bugreport.sh
+source "$repo_root/usr/bin/cachyos-bugreport.sh"
+original_collect_sensitive_values=$(declare -f collect_sensitive_values)
+
+test_dir=$(mktemp -d)
+trap 'rm -rf "$test_dir"' EXIT
+
+collect_sensitive_values() {
+    cat <<'EOF'
+<home-dir-redacted>	/home/alex
+<username-redacted>	alex
+<ip-address-redacted>	192.0.2.44
+<ip-address-redacted>	2001:db8::44
+<mac-address-redacted>	02:11:22:33:44:55
+<machine-id-redacted>	0123456789abcdef0123456789abcdef
+<uuid-redacted>	ABCD-1234
+<ssid-redacted>	cachyos
+<usb-serial-redacted>	CURRENT-USB-123
+EOF
+}
+
+LOG_FILENAME="$test_dir/report.log"
+cat >"$LOG_FILENAME" <<'EOF'
+uname: Linux cachyos 7.1.6-1-cachyos #1 SMP PREEMPT_DYNAMIC x86_64 GNU/Linux
+repo cachyos-v4 package linux-cachyos kernel 7.1.6-1-cachyos
+firmware version 6.18.44.1
+Host Name: cachyos
+systemd: Set hostname to cachyos.
+home=/home/alex/config user=alex allocation=ok
+current IPv4=192.0.2.44 current IPv6=2001:db8::44
+MAC=02:11:22:33:44:55 machine=0123456789abcdef0123456789abcdef
+filesystem UUID=ABCD-1234 standard=123e4567-e89b-12d3-a456-426614174000 historical PARTUUID=DEAD-BEEF
+connected to cachyos; SerialNumber: CURRENT-USB-123
+old firewall SRC=198.51.100.22 DST=2001:db8::99 MAC=AA:BB:CC:DD:EE:FF
+old lease address=10.2.3.4 gateway=2001:db8::1
+old wifi SSID="Old Cafe" and access point 'Older Cafe'
+old device Serial Number: OLD-USB-456
+machine-id=abcdefabcdefabcdefabcdefabcdefab
+opened QUrl("file:///mnt/private/one") then QUrl("file:///home/alex/two") safely
+contact maintainer@example.org
+EOF
+
+redact >/dev/null
+
+cat >"$test_dir/expected.log" <<'EOF'
+uname: Linux <hostname-redacted> 7.1.6-1-cachyos #1 SMP PREEMPT_DYNAMIC x86_64 GNU/Linux
+repo cachyos-v4 package linux-cachyos kernel 7.1.6-1-cachyos
+firmware version 6.18.44.1
+Host Name: <hostname-redacted>
+systemd: Set hostname to <hostname-redacted>.
+home=<home-dir-redacted>/config user=<username-redacted> allocation=ok
+current IPv4=<ip-address-redacted> current IPv6=<ip-address-redacted>
+MAC=<mac-address-redacted> machine=<machine-id-redacted>
+filesystem UUID=<uuid-redacted> standard=<uuid-redacted> historical PARTUUID=<uuid-redacted>
+connected to <ssid-redacted>; SerialNumber: <usb-serial-redacted>
+old firewall SRC=<ip-address-redacted> DST=<ip-address-redacted> MAC=<mac-address-redacted>
+old lease address=<ip-address-redacted> gateway=<ip-address-redacted>
+old wifi SSID=<ssid-redacted> and access point '<ssid-redacted>'
+old device Serial Number: <usb-serial-redacted>
+machine-id=<machine-id-redacted>
+opened QUrl("file://<path-redacted>") then QUrl("file://<path-redacted>") safely
+contact <email-address-redacted>
+EOF
+
+diff -u "$test_dir/expected.log" "$LOG_FILENAME"
+
+eval "$original_collect_sensitive_values"
+SUDO_USER=root
+ip() {
+    printf '%s\n' '2: eth0 inet 192.0.2.55/24 scope global eth0' '2: eth0 inet6 2001:db8::55/64 scope global'
+}
+lsblk() { printf '%s\n' 'MOCK-UUID'; }
+nmcli() { printf '%s\n' '802-11-wireless:Cafe:Lab'; }
+udevadm() { printf '%s\n' 'E: ID_SERIAL_SHORT=MOCK-USB'; }
+inventory=$(collect_sensitive_values)
+grep -Fxq $'<ip-address-redacted>\t192.0.2.55' <<<"$inventory"
+grep -Fxq $'<ip-address-redacted>\t2001:db8::55' <<<"$inventory"
+grep -Fxq $'<uuid-redacted>\tMOCK-UUID' <<<"$inventory"
+grep -Fxq $'<ssid-redacted>\tCafe:Lab' <<<"$inventory"
+grep -Fxq $'<usb-serial-redacted>\tMOCK-USB' <<<"$inventory"
+
+collect_sensitive_values() { :; }
+LOG_FILENAME="$test_dir/empty-inventory.log"
+printf '%s\n' 'plain diagnostic text remains intact' >"$LOG_FILENAME"
+redact >/dev/null
+grep -Fxq 'plain diagnostic text remains intact' "$LOG_FILENAME"
+
+printf '%s\n' 'bugreport redaction tests: PASS'
--- a/tests/test-bugreport-redaction.sh
+++ b/tests/test-bugreport-redaction.sh
@@ -10,7 +10,7 @@ original_collect_sensitive_values=$(declare -f collect_sensitive_values)
 test_dir=$(mktemp -d)
 trap 'rm -rf "$test_dir"' EXIT
 
-# --- Test 1: Main redaction pass covering all sensitive data classes ---
+# --- Test 1: Main redaction pass covering all sensitive data classes & edge cases ---
 collect_sensitive_values() {
     cat <<'EOF'
 <home-dir-redacted>	/home/alex
@@ -55,8 +55,10 @@ pcieport 0000:00:1c.4: AER: Corrected error received
 NetworkManager: device eth0 connected; carrier on
 CPU0: Thermal 100 C
 AA:BB:CC:DD:EE:FF 11:22:33:44:55:66 22:33:44:55:66:77
-nfs: server 192.168.1.50 not responding
+route fe80:00:11:22:33:44:55:66 metric 100
+nfs: server 198.51.100.22 not responding
 Failed to connect to 10.0.0.5:8080
+WireGuard endpoint 203.0.113.5:51820
 EOF
 
 redact >/dev/null
@@ -84,16 +86,18 @@ xhci_hcd 0000:0e:00.0: xHCI Host Controller
 nvme nvme0: pci function 0000:10:00.0
 amdgpu 0000:7a:00.3: amdgpu: Fetched VBIOS from VFCT
 pcieport 0000:00:1c.4: AER: Corrected error received
-NetworkManager: device eth0 connected; carrier <ssid-redacted>
+NetworkManager: device eth0 connected; carrier on
 CPU0: Thermal 100 C
 <mac-address-redacted> <mac-address-redacted> <mac-address-redacted>
+route fe80:00:11:22:33:44:55:66 metric 100
 nfs: server <ip-address-redacted> not responding
 Failed to connect to <ip-address-redacted>
+WireGuard endpoint <ip-address-redacted>
 EOF
 
 diff -u "$test_dir/expected.log" "$LOG_FILENAME"
 
-# --- Test 2: Custom hostname (petes-laptop) redaction across non-fixed positions ---
+# --- Test 2: Custom hostname (petes-laptop) and Avahi conflict suffix (petes-laptop-2) ---
 collect_sensitive_values() {
     cat <<'EOF'
 <hostname-redacted>	petes-laptop
@@ -112,17 +116,17 @@ redact >/dev/null
 
 cat >"$test_dir/hostname-custom-expected.log" <<'EOF'
 NetworkManager[689]: <info> hostname changed from "localhost" to "<hostname-redacted>"
-avahi-daemon[702]: Host name conflict, retrying with petes-laptop-2
+avahi-daemon[702]: Host name conflict, retrying with <hostname-redacted>
 kernel: Linux version 7.1.6-1-cachyos (linux-cachyos@<hostname-redacted>) #1
 dbus-daemon: [session uid=1000 pid=900] on <hostname-redacted>
 EOF
 
 diff -u "$test_dir/hostname-custom-expected.log" "$LOG_FILENAME"
 
-# --- Test 3: Collection mock validations (filtering PCI BDF, nmcli SSID, user fallback) ---
+# --- Test 3: Collection mock validations (filtering PCI BDF, date/timestamp serials, nmcli contract, user detection) ---
 eval "$original_collect_sensitive_values"
 
-# Test udevadm PCI address / generic serial filtering
+# Test udevadm PCI address / generic / timestamp serial filtering
 SUDO_USER=root
 udevadm() {
     printf '%s\n' \
@@ -130,8 +134,12 @@ udevadm() {
         'E: ID_SERIAL_SHORT=0000:10:00.0' \
         'E: ID_SERIAL_SHORT=000000000' \
         'E: ID_SERIAL_SHORT=9876543210' \
+        'E: ID_SERIAL_SHORT=1234567890' \
         'E: ID_SERIAL_SHORT=0' \
-        'E: ID_SERIAL_SHORT=VALID-USB-DEVICE'
+        'E: ID_SERIAL_SHORT=202404073115' \
+        'E: ID_SERIAL_SHORT=20231122' \
+        'E: ID_SERIAL_SHORT=VALID-USB-DEVICE' \
+        'E: ID_SERIAL_SHORT=SN12345678'
 }
 ip() { :; }
 lsblk() { :; }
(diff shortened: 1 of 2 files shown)


petejm:
Reviewed this against head `27c8943` on GNU bash 5.3.15 and GNU sed 4.10. It is a draft, so
take the blockers as "before this leaves draft" rather than as objections to the approach. The
enumerate-then-redact architecture is the right call and the diagnosis of the current
false-positive problem is correct and independently reproducible.

Things worth saying first, because they are the parts I would not want lost in a rewrite:

* It ships a test. Deterministic, it genuinely passes here, and it is the first test covering
  this path at all.
* The `[ "${BASH_SOURCE[0]}" = "$0" ]` guard is the enabling change that makes the test
  possible, since the script becomes sourceable without triggering `check_root` and `upload`.
* MAC and machine-id are read from sysfs and `/etc/machine-id` directly rather than by shelling
  out, so the two highest-value classes still work on a headless or minimal system. I checked
  degradation with `ip`, `lsblk`, `nmcli` and `udevadm` all absent from `PATH`: clean exit 0,
  log intact.
* The empty-value guard in `emit_sensitive_values` is load-bearing rather than defensive.
  `lsblk -rno UUID` emitted four blank lines on my machine, and each would otherwise have
  become an empty pattern matching at every position.
* Emitting the home directory before the username is the right order, so `/home/alex` becomes
  `<home-dir-redacted>` rather than `/home/<username-redacted>`.
* Both headline claims verify independently of your fixtures: `6.18.44.1` survives, and
  `linux-cachyos`, `cachyos-v4` and `7.1.6-1-cachyos` are preserved.

## 1. `ID_SERIAL_SHORT` enumeration picks up PCI bus addresses

`udevadm info --export-db` is swept across the whole device database, not just USB devices, and
every xHCI root hub reports its parent controller's PCI address in that field. Real values from
my machine:

```
E: ID_SERIAL_SHORT=0000:0e:00.0
E: ID_SERIAL_SHORT=0000:10:00.0
E: ID_SERIAL_SHORT=0000:7a:00.3
```

Those reach the unanchored default substitution, and the result on a realistic dmesg excerpt is:

```
xhci_hcd <usb-serial-redacted>: xHCI Host Controller
nvme nvme0: pci function <usb-serial-redacted>
amdgpu <usb-serial-redacted>: amdgpu: Fetched VBIOS from VFCT
pcieport <usb-serial-redacted>: AER: Corrected error received
```

Which GPU, which NVMe and which port logged the AER are all gone. This is the same
over-redaction the PR sets out to remove, reached by a different route, and it will fire on
nearly every machine. The same sweep also collected `000000000`, `9876543210` and
`202404073115` here, which are generic enough to be shared across cheap devices.

## 2. Enumerated literals are substituted without boundary anchoring

The username substitution anchors with `\b` and the SSID one is gated to a line context, but the
default branch that handles home directory, IP, MAC, machine-id, UUID and USB serial has
neither. Any short or common enumerated value then rewrites unrelated text.

Two-character SSID `on`, through the gated branch:

```
NetworkManager: device eth0 c<ssid-redacted>nected; carrier <ssid-redacted>
```

USB serial `0`, through the ungated branch:

```
CPU<usb-serial-redacted>: Thermal 1<usb-serial-redacted><usb-serial-redacted> C
```

Requiring a non-alphanumeric boundary on enumerated literals, plus a minimum length and a
denylist for PCI-address-shaped and all-zero values, would close both this and point 1 without
disturbing the design.

## 3. Netfilter `MAC=` fields are no longer redacted

The `[^[:xdigit:]:]` guards exclude any MAC embedded in a longer colon-hex run, and the kernel
netfilter log `MAC=` field is exactly that: a 14-byte chain carrying both destination and source
MAC. Same input, both versions:

```
this PR:  MAC=00:11:22:33:44:55:66:77:88:99:aa:bb:08:00 SRC=<ip-address-redacted>
current:  MAC=<mac-address-redacted>:<mac-address-redacted>:08:00 SRC=<ipv4-redacted>
```

The enumerated pass covers the local NIC's own MAC, so what leaks here is the peer or gateway
MAC, which is the case the previous-boot fallback exists for.

Related, same cause: because `g` resumes after the consumed trailing character, adjacent MACs
separated by one delimiter lose the middle match.

```
input:    AA:BB:CC:DD:EE:FF 11:22:33:44:55:66 22:33:44:55:66:77
this PR:  <mac-address-redacted> 11:22:33:44:55:66 <mac-address-redacted>
```

This is the same consume-the-separator hazard the description says was fixed for `file://`
URLs; it just was not applied here.

## 4. Dropping the blanket IPv4 pattern trades a false-positive class for a false-negative one

Coverage is now current addresses plus `SRC=`/`DST=` plus the
`ip_address|address|gateway|nameserver` fields. Anything outside that is untouched, for example
`nfs: server 192.168.1.50 not responding`, `Failed to connect to 10.0.0.5:8080`, WireGuard
endpoints and any address belonging to a previous network. The current pattern catches all of
those.

Worth naming that trade explicitly rather than treating it as strictly better. The motivating
false positive is real, and broader than reported:

```
$ printf 'firmware 6.18.44.1 and bcdDevice 1.02.03.04\n' \
  | sed 's/\b\([0-9]\{1,3\}\.\)\{3\}[0-9]\{1,3\}\b/<ipv4-redacted>/g'
firmware <ipv4-redacted> and bcdDevice <ipv4-redacted>
```

But `6.18.44.1` is a syntactically valid dotted quad, so no structural guard separates the two.
The discriminator has to be line context such as `version|firmware|revision|bcdDevice`, which
would let the blanket pattern stay rather than be dropped.

## 5. Positional hostname redaction misses most occurrences

Restricting to three fixed positions leaves these, on hostname `petes-laptop`:

```
NetworkManager[689]: <info> hostname changed from "localhost" to "petes-laptop"
avahi-daemon[702]: Host name conflict, retrying with petes-laptop-2
kernel: Linux version 7.1.6-1-cachyos (linux-cachyos@petes-laptop) #1
dbus-daemon: [session uid=1000 pid=900] on petes-laptop
```

The motivation is sound, since with the default hostname `cachyos` a `\b`-anchored global
replace does match inside `linux-cachyos` because `-` is a non-word character. A middle path is
to keep the global replace but skip it when the match is preceded by `-` or `/`, or when the
line matches `linux-cachyos|cachyos-v[34]|cachyos-znver4`.

The `--no-hostname` addition to both `journalctl` calls is the right fix at the source, and
handling systemd's angle-bracket `Set hostname to` form is a good catch.

## 6. The nmcli connection name is not the SSID

`connection show` returns the profile NAME. NetworkManager stores the SSID in
`802-11-wireless.ssid`; the name defaults to it but diverges on a renamed profile or an
auto-suffixed duplicate such as `MySSID 1`. Reading `-g 802-11-wireless.ssid` per connection is
exact. `iwd`-only systems are missed entirely either way.

Using `--escape no` to preserve colons inside the name is a nice touch, and the `Cafe:Lab`
fixture shows it was deliberate.

## Smaller notes

* A malformed expression makes `sed -Ei` fail before touching the file and `set -e` then aborts,
  which is the safe direction since `upload` is never reached. But the unredacted
  `cachyos-bugreport.log` is left in the working directory, and a user told to run the script
  and attach the log will attach it. A `trap` removing or renaming the log on non-zero exit
  would close that.
* `sed_escape` is not declared `local`, so it leaks into the caller's namespace.
* `\b((PART)?UUID=)` will not match `ID_FS_UUID=`, since `_` is a word character and suppresses
  the boundary.
* User detection is `SUDO_USER` only, so `pkexec`, `su -` and a direct root login all skip
  username and home-dir redaction. Pre-existing rather than introduced here, but cheap to add
  `logname` or `PKEXEC_UID` fallbacks while this area is being reworked.

Points 1 through 3 look cheap to fix and none of them disturb the architecture. Happy to send
patches for any of them if that is useful.

_Disclosure: Researched and drafted by an AI agent (Claude), guided & reviewed by @petejm_

alexsmolya:
Thanks for the thorough review and edge-case feedback @petejm!

I have updated the branch with fixes across the identified areas:
- **SSID handling**: Refactored boundary matching to handle punctuation-containing SSIDs (e.g. `Lab!`, `-foo-`, `Cafe:Lab`) and consecutive mentions without over-redacting common prose words.
- **IPv4 validation & version preservation**: Replaced whole-line skipping with contextual token matching. Endpoints on mixed lines are redacted while dotted kernel/firmware/diagnostic version strings are preserved, and invalid octets are rejected.
- **MAC boundaries & label variants**: Added support for mixed-case and prefixed labels (e.g. `Mac:`, `wifi_mac:`, `client_hwaddr=`) while strictly protecting longer colon-hex sequences and IPv6 route diagnostics.
- **Test coverage**: Expanded the test suite and verification mocks to enforce full invocation contracts and cover these edge cases.

petejm:
_Disclosure: researched and drafted by an AI agent (Claude), guided and reviewed by @petejm.
The commands and test runs described below were executed; where a result comes from a stub or a
constructed fixture rather than real hardware, I say so at that point._

Re-reviewed against head `cf94eb3` on GNU bash 5.3.15 and GNU sed 4.10, same versions as last
time so the results are directly comparable. Your test suite passes here.

All six points from the previous round are genuinely addressed, and I checked each one rather
than taking the summary on trust:

1. **USB serials.** The `grep -Ev` filter removes every PCI address (bus:device.function). On my
   own device database, 13 `ID_SERIAL_SHORT` entries across 5 distinct values, 4 of them
   PCI-address-shaped, and all 4 are filtered. Zero survive. That is one machine's device set
   rather than a general proof of coverage, so treat it as a spot check.
2. **Boundary anchoring.** The minimum length of 3 plus the anchored branches close the `0` and
   `on` cases.
3. **Netfilter MAC chains.** `MAC=` now redacts to
   `<mac-address-redacted>:<mac-address-redacted>:08:00`. The doubled expressions also fix the
   consume-the-separator gap: I tested runs of three and four whitespace-separated MACs and all
   of them are redacted, which is the right result for the reason you would want, since a second
   pass over the same line catches whatever the first pass skipped. Runs of five and six are
   clean too.
4. **IPv4.** The blanket pattern is back with a line-context guard. See the finding below.
5. **Hostname.** The global replace with `[^a-zA-Z0-9_/-]` boundaries handles all four of my
   cases, including `my-laptop-2` and `linux-cachyos@my-laptop`, while preserving
   `linux-cachyos` and `cachyos-v4`.
6. **SSID.** Reading `-g 802-11-wireless.ssid` per connection is exact, and the mock enforces the
   argument contract.

The four smaller notes are all handled too, including `ID_FS_UUID=` and the `PKEXEC_UID` and
`logname` fallbacks.

Two findings below. The first is a blocker: it is a consequence of the architecture rather than
of any single expression, and it can hand a user a report they have been told is redacted when
it is not. The second is lower severity and I have not caught it firing on my hardware, but it
is in scope for this PR because it is the exact over-redaction class the PR sets out to remove.
I explain below why it does not currently fire.

## 1. A failure inside `collect_sensitive_values` silently disables the rest of the redaction

`redact()` reads the inventory through process substitution:

```bash
done < <(collect_sensitive_values)
```

The exit status of a process substitution is never observed, and `set -e` cannot see it. If the
collector dies partway, `redact()` proceeds with whatever it received, `sed` succeeds,
`CLEANUP_ON_ERROR` is set to 0, and the script exits 0 with the log on disk and `upload()`
offering to publish it.

I tested four failure modes against the real script semantics. All four behave identically:

| Failure mode | Script exit | Log kept | SSID redacted |
|---|---|---|---|
| collector exits 1 midway | 0 | yes | **no** |
| collector fails immediately | 0 | yes | **no** |
| collector killed by SIGTERM | 0 | yes | **no** |
| collector returns 143 | 0 | yes | **no** |

The hostname still gets redacted in all four, because the positional and labeled hostname
expressions are static and do not depend on the inventory. That is what makes this hard to
notice: the report looks redacted.

**This is reachable without anything going wrong.** `emit_sensitive_values` ends each iteration
with an AND list:

```bash
[ -n "$value" ] && [ "${#value}" -ge 3 ] && printf '%s\t%s\n' "$replacement" "$value"
```

When the **last** line it reads is empty or shorter than 3 characters, that list returns 1, so
the function returns 1, so the pipeline returns 1, and `set -e` kills the collector subshell at
that stage. Every class enumerated *after* it is skipped.

Confirmed with the collector stubbed so that only the UUID stage varies:

```
UUID list ends with a value        -> hostname, MAC, machine-id, UUID, SSID, serial   (exit 0)
UUID list ends with a blank line   -> hostname, MAC, machine-id, UUID                 (aborts)
blank line in the MIDDLE of list   -> all classes present                             (exit 0)
UUID list is one blank line only   -> hostname, MAC, machine-id                       (aborts)
```

So the trigger is not "a blank value exists", it is "the last value of a class is blank", which
makes it depend on enumeration order and therefore on hardware. Worth noting that genuinely
empty output does not abort, since a `while` loop that never executes its body returns 0. It
takes a blank *line*, which is exactly what `lsblk -rno UUID` emits.

Two ordinary ways to land on it:

* `lsblk -rno UUID` emits an empty line for every whole-disk device and every unformatted or
  unpartitioned device. On my machine there is such a line (the `nvme0n1` disk itself) and it
  happens not to be last, so collection survives by luck. A second unformatted drive, a USB
  stick with no filesystem, or an optical drive enumerated last puts the blank at the end.
* The MAC stage is worse, because a TUN interface has a 1 byte `address` file containing only a
  newline, which reaches `emit_sensitive_values` as an empty line. On my machine:

  ```
  lo            18 bytes   -> deleted by the 00:00:00:00:00:00 sed
  tailscale0     1 byte    -> emits an EMPTY line
  wlan0         18 bytes   -> real MAC
  ```

  `wlan0` sorts last here so the stage returns 0. Reorder that and it does not. I verified the
  ordering dependence against a constructed `/sys/class/net` tree:

  ```
  enp3s0, lo, wg0   -> emit exit = 1, collection aborts
  ```

  Enumeration is a plain bash glob (`/sys/class/net/*/address`), so the ordering is the shell's
  own sort with no external command involved, and it is alphabetical under both the default
  locale and `LC_ALL=C`. `wg0`, `tun0` and ZeroTier's `zt*` all sort after `enp*`, `eth*` and
  `lo`, and `zt*` sorts after `wlan0` as well.

I should be straight about what I have and have not seen. I have not triggered this end to end
on real hardware. On two CachyOS machines the precondition is present but does not land last:
`lsblk -rno UUID` emits a blank line on both, and a `tailscale0` interface emits an empty one,
but a real MAC sorts after them each time. So the mechanism is confirmed at the function level
and the enumeration-order argument holds; what I do not have is a machine that happens to
enumerate in the losing order. I am raising it as a blocker anyway because the precondition is
already present on both machines I can check, and the thing standing between that and a leaked
report is which interface happens to sort last.

When it fires, UUIDs, SSIDs and USB serials are never collected, and the user is offered an
upload of a report they have been told is redacted.

The fix is small and turns this fail-closed:

```bash
inventory="$(collect_sensitive_values)" || {
    echo "ERROR: could not enumerate values to redact; refusing to continue." >&2
    exit 1
}
...
done <<<"$inventory"
```

and making the emit loop not return the status of its last test, for example by ending the loop
body with `:` or using an explicit `if`.

Both are worth doing, though not quite symmetrically. The loop change alone leaves the first
hole fully open: an external death of the collector still proceeds silently with a partial log.
The `|| exit 1` alone does catch all four death modes above, and last-stage failures too, since
those become the function's own return value. Mid-stream stage failures survive it, but only
because bash does not inherit `errexit` into `$(...)` unless `inherit_errexit` is set, which
this script does not do. Applying both is the version that does not rest on that detail.

## 2. Package versions that are valid dotted quads are redacted as IP addresses

The blanket IPv4 rule is skipped only when the line also matches
`version|release|kernel|bcdDevice|revision`. All eight occurrences of `6.18.44.1` in the suite
carry one of those words, and no fixture anywhere is a bare valid dotted quad with no label and
no guard word, so the unguarded path is never exercised at all:

```
firmware version 6.18.44.1 loaded    -> preserved
firmware 6.18.44.1 loaded            -> firmware <ip-address-redacted> loaded
firmware: 6.18.44.1 loaded           -> firmware: <ip-address-redacted> loaded
6.18.44.1                            -> <ip-address-redacted>
```

The Installed Packages section is a candidate for this, though on my machine it happens not to
fire, for a reason worth knowing about. The section is built with
`pacman -Ss | grep --color=never "^cachyos-v4/.*\[installed\]" || true`, and on a v4 install
the optimized packages overwhelmingly come
from `cachyos-extra-v4` rather than `cachyos-v4`. Here that is 40 packages captured against 972
in the sibling Cachy repos. None of the 40 carry a dotted-quad version, so nothing is
over-redacted in practice.

The exposure is in what the section is presumably meant to include. Of my 760 installed
`cachyos-extra-v4` packages, 7 have a version that is a valid dotted quad:

```
alsa-lib 1.2.16.1-1.1          libldac 2.0.2.6-1.1
gcr-4 4.4.0.1-1.1              libmodplug 0.8.9.0-7.1
imagemagick 7.1.2.29-2.1       socat 1.8.1.3-1.1
kquickimageeditor 0.6.2.1-2.1
```

Widen the grep and every one of those becomes `<ip-address-redacted>`. Flagging the repo-prefix
gap separately since it is pre-existing and not this PR's doing.

The underlying gap, not the packages case specifically, is the same over-redaction class the PR
sets out to remove, so it seems worth closing rather than accepting. Adding `firmware` to the
guard alternation is not enough on its own,
because the bare `6.18.44.1` case has no label at all. A structural discriminator would help
more: a dotted quad immediately followed by `-` and a pacman release number is not an address.
That only covers the Installed Packages shape, though. Restricting the blanket rule to lines
outside the Installed Packages block is the one that actually closes the bare case, and it is
worth noting your own suite already encodes the boundary this relies on, since it asserts
`address=999.999.999.999` survives.

Note also that `firmware` currently sits in the labeled-IP alternation on the
`ip_address|address|gateway|...` line, so `firmware: <quad>` is redacted by that expression
before the guarded blanket rule is ever consulted.

## Smaller notes

* Bare IPv6 is not redacted, while bare IPv4 now is. `peer 2001:db8:85a3::8a2e:370:7334 timed
  out` and `route via fe80::1234:5678:9abc:def0` both survive, and the `[^0-9A-Fa-f:]` guards on
  the MAC rules deliberately protect longer colon-hex runs. The enumerated pass covers this
  machine's own addresses, so what leaks is peer and previous-network IPv6. Worth naming
  explicitly given @azdanov's point on #250 that IPv6 can be identifiable.
* `address: aa:bb:cc:dd:ee:ff` becomes `address: <ip-address-redacted>`, because the labeled
  IPv6 expression runs before the labeled MAC expression and matches a MAC as colon-hex.
  `gateway:` behaves the same way, and so will any other label the two alternations share.
  Reversing the order of the two expressions flips the label, which is what confirms the cause.
  The value is redacted either way, so this is only a labeling loss.
* `^(.)\1+$` in the serial filter is a backreference inside an ERE, which is a GNU extension
  rather than POSIX. GNU grep is what CachyOS ships so this works, but it does fail on stricter
  implementations, and it fails by silently matching less rather than by erroring. That direction
  is the harmless one, since it only lets a repeated-character serial into the redaction list.
  Only worth knowing about.
* The minimum length of 3 applies to the inventory, so a 2 character hostname or SSID never
  reaches the dynamic rules and is not redacted in free text. The static expressions are not
  length gated and still catch the labeled and positional cases, so `uname:`, `hostname:`,
  `Host Name:` and `SSID:` are all covered regardless of length. What escapes is a bare two
  character occurrence somewhere else in the log. Probably the right trade against the `on`
  and `0` false positives, but it is a deliberate gap rather than an oversight, so it may
  deserve a comment.
* On a redaction failure the trap deletes the report entirely, with no message explaining why
  the file the user was told to attach is missing. Renaming it to `.unredacted` and saying so
  would be friendlier, and still keeps it out of an accidental upload.

Point 1 I would want closed before merge. Point 2 I am not asking you to treat as blocking,
since I have not observed it firing live, but I would rather it were tracked in a linked issue
than dropped, because it is the class this PR exists to remove. Happy to send patches for
either, or for the smaller notes, whichever is more useful to you.

petejm:
_Disclosure: posted by an AI agent (Claude) on behalf of @petejm, who reviewed and approved this comment._

Process note rather than a review point, so feel free to ignore if this is already known.

#252 is also open against the redaction code in `usr/bin/cachyos-bugreport.sh`, and also
references #250. It takes a narrower approach: per @drinkingoutofcups on #250, an exception for
the hostname when it is "cachyos", rather than the enumerate-then-redact rework here.

The two will conflict, since both edit the same expressions in the same file. I have reviewed
both and have no stake in which one lands. Flagging it so whoever merges gets to make that
choice deliberately rather than meet it as a merge conflict. cc @drinkingoutofcups

alexsmolya:
Thanks for catching the failure-observability issue. `collect_sensitive_values` is now captured in a checked command substitution, so immediate or partial collection failure aborts redaction instead of continuing with an incomplete inventory. `emit_sensitive_values` also no longer fails just because its final candidate is blank or too short. Regression tests cover immediate failure, partial-output failure, cleanup/no unsafe continuation, and trailing blank/short candidates. I intentionally kept the dotted-quad/package-version concern out of this blocker fix; it should be handled separately as a focused follow-up to avoid broadening scope.

alexsmolya:
I'm continuing work on this as a partner, not as a competitor. The existing implementation and review history contain valuable evidence, and I want to preserve that work while helping bring the redaction behavior to a reliable, fail-closed conclusion.

kvnloo:
Exact head 2f5297aaf5e2cecdfa6c454d849e87d0a4e049d1 (`cachyos-bugreport: fail closed on inventory errors`), not the first commit in the PR.

On this clone, `bash tests/test-bugreport-redaction.sh` printed:

    ERROR: could not enumerate values to redact; refusing to continue.
    ERROR: could not enumerate values to redact; refusing to continue.
    bugreport redaction tests: PASS

Exit 0. Log sha256 `400a700f8e52629b57e36e8233e1453211a6dcd8a6d43d29409d9c1eda838603`.

The two ERROR lines are from the fail-closed inventory path the tip added; the script still passes. I did not generate a real `cachyos-bugreport` tarball on this machine because that would include host identifiers. I only ran the in-tree test harness.

ventureoo:
I'm not strictly against AI contributions, but this one is just painful to review. Please make it easier to read.

PR #257, opened 2026-08-20
udev: add rule to keep USB HID input devices out of runtime PM

VECTORG99:
## Summary

Adds `usr/lib/udev/rules.d/75-usb-hid-pm.rules`, which disables USB runtime PM (autosuspend) for **external, removable** HID input devices so they stay fully awake while in use.

## Rationale

When USB autosuspend engages on HID input devices, the first input event after the idle timeout is **silently dropped** — the kernel USB PM documentation explicitly warns about this for usbhid: *"typing on a suspended keyboard will frequently result in lost keystrokes"*, and mice may not issue a remote-wakeup on motion at all. In some cases the device resumes into a broken state and requires a USB reset to recover.

This is a **correctness** issue, not a performance tuning. CachyOS already disables runtime PM for audio (`20-audio-pm.rules`) and disk devices (`50-sata.rules`, `69-hdparm.rules`) for similar reasons, so keeping external HID input devices out of runtime PM is consistent with that stance.

The risk is real on any system where `power/control` gets flipped to `auto` — which happens under battery profiles, `tlp`, `power-profiles-daemon`, or a globally enabled `usbcore.autosuspend` kernel parameter. On such systems the existing `modprobe.d` workaround is also ineffective because `usbcore` is built into the CachyOS kernel (`CONFIG_USB=y`), so `modprobe.d` options are silently ignored (see issue #191).

## Scope

The rule matches **only external, removable HID interfaces**:

- `ATTR{bInterfaceClass}=="03"` — HID class, matched on the USB **interface** device.
- `ATTRS{removable}=="removable"` — only devices whose USB parent is flagged removable. Internal keyboards/touchpads report `fixed` (or `unknown`) and **stay on autosuspend**.
- `TEST=="../power/control"` — guard for hosts/controllers without a `power/control` attribute; the rule is a no-op there.
- `ATTR{../power/control}="on"` — writes the parent USB device's runtime PM control (the interface itself has no `power/control`; the `../` relative path is the same mechanism used by upstream udev's `42-usb-hid-pm.rules`).

Formatting follows the repo's existing multi-line backslash style (see `71-nvidia.rules`, `50-sata.rules`).

## Verification

- Attribute values verified on real hardware: a Holtek `04D9:FC5F` mouse exposes `bInterfaceClass=03` on its interface and `removable="removable"` on the USB device.
- The `../` relative-path mechanism was cross-checked against the upstream udev reference rule (`42-usb-hid-pm.rules`) and `udevadm verify` passes on the rule file.
- A full end-to-end `udevadm test --action=add` on target hardware is still recommended in CI before merge (it requires the rule to be installed in a `rules.d` directory).


kvnloo:
Exact head 03bfc64552df4599b47df22eddb8f555044d62ba on CachyOS 7.2.2-1-cachyos. I did not install `75-usb-hid-pm.rules`; this is live sysfs on the current distro ruleset.

Removable USB HID already has parent `power/control=on` (`runtime_status=active`, `autosuspend=2`):

- 1-2 Zaunkoenig M2K `0483:a3cf`, removable; interface class 03
- 1-3 Wooting 60HE `31e3:1312`, removable; interfaces 1.0–1.4 class 03
- 1-7 Nintendo WUP-028 `057e:0337`, removable; interface class 03

So on this desktop the proposed `ATTR{../power/control}="on"` would be a no-op for the devices I care about. I cannot claim a latency win.

Rule-match concern: `ATTR{bInterfaceClass}=="03"` is an interface attribute. Combined with `TEST=="../power/control"` it is aimed at `usb_interface` add events, not `usb_device`. If udev evaluates this against the device node, `bInterfaceClass` is empty and the rule never fires. The file also lacks a trailing newline.

Asus `0b05:18f3` AURA LED is `removable=fixed` with a class-03 interface; the `ATTRS{removable}=="removable"` filter correctly skips it on this box.

VECTORG99:
Thanks for the review and the detailed hardware check @kvnloo!

I've updated the rule to address both concerns:

1. **Explicit interface match**: Added `ENV{DEVTYPE}=="usb_interface"` so udev only evaluates the rule on interface add events where `bInterfaceClass` is defined, eliminating any ambiguity against `usb_device` evaluation.
2. **Formatting**: Added the missing trailing newline.

Regarding the rationale / no-op on desktop setups:
While standard desktop configurations on AC power often keep USB HID devices at `power/control="on"` by default, this rule is primarily protective for:
- **Laptops / Battery profiles**: Power management tools (e.g. `power-profiles-daemon`, `tlp`, `powertop --auto-tune`) or battery power profiles commonly switch all USB devices to `power/control="auto"`.
- **Global autosuspend configurations**: On systems where autosuspend is globally enabled (`usbcore.autosuspend=2`), external input peripherals can be suspended after idle timeouts, introducing wake-up/poll latency spikes on initial movement.

Having this rule guarantees that external HID input devices consistently stay awake regardless of active power profiles or battery status, aligning with CachyOS's performance-oriented tuning (similar to the audio and disk PM rules).

Verified locally with `udevadm verify` and `udevadm test` against target HID interface nodes.

damachine:
> High-polling-rate input devices (gaming mice, fast keyboards, etc.) can suffer from measurable input latency and occasional missed events when their USB link autosuspend engages. 

Hey! I don't quite understand, but what kind of input latency improvement did you observe, and how was it measured?

VECTORG99:
Hey! You're right to question that — I framed it poorly. The issue isn't really about measurable latency improvements in the performance sense. The actual problem is **correctness**: when autosuspend engages on HID devices, the first input after the idle timeout is silently dropped (the kernel docs explicitly warn about this for usbhid), and in some cases the device resumes into a broken state requiring a USB reset to recover — which I reproduced on my own hardware during testing.

The "measurable input latency" wording was imprecise. I'll update the PR description to reflect the real rationale: preventing dropped inputs and broken resume states, consistent with how CachyOS already treats audio and disk PM.

PR #260, opened 2026-09-06
systemd: recover Plymouth after quit timeout

sheno4:
On my three-monitor setup, boot could occasionally get stuck at the CachyOS splash or a black screen. In the affected boots, `plymouth-quit.service` timed out while `plymouthd` remained alive and continued holding DRM master, preventing the display manager and KWin from accessing the GPU.

This change adds recovery handling for that failure path. If `plymouth-quit.service` times out, the remaining `plymouthd` process is terminated and systemd waits for the cleanup to complete before the display manager proceeds.

Normal boots are unaffected and continue to use the Plymouth splash as usual.

The fix has now been validated on the affected system and successfully recovers from the previously reproducible boot failure.

Related to CachyOS/linux-cachyos#940.


ventureoo:
Honestly, I don't find such workarounds acceptable. Can you plase try this version of plymouth instead? https://github.com/CachyOS/distribution/issues/480#issuecomment-5453391470

PR #267, opened 2026-09-15
Handle xswap configuration

### LINKED ISSUE #266: Migrate from zram to xswap
There will be a virtual swap soon available, which fixes the drawbacks of zram. As soon its merged upstream we can start the migration on CachyOS-Settings.

The latest kernel will then support zram and LTS will not have it. So likely we need to make it conditional.

https://lore.kernel.org/lkml/20260913075014.1732524-1-hebaoquan@kylinos.cn/

ventureoo:
Closes: https://github.com/CachyOS/CachyOS-Settings/issues/266


ventureoo:
v2: Fix generator double-triggering on systemctl daemon-reload.

